Issue #3, opened 2016-03-31
Develop a process for Working Group formation


caniszczyk:
There is an interest to form Working Groups around specific topics.

We should have a simple formal process to establish these.

jonboulle:
Can you give some examples of topics for which there's been interest in
creating working groups?

chriswright:
@jonboulle some topics that came up:
- Service Broker
- Storage
- Networking
- OpenStack collaboration

jonboulle:
Thanks - and what's the thinking in terms of how these working groups
interact with CNCF projects? I'm just trying to understand what the
implications are

caniszczyk:
drafted a proposal here in google doc form for comment tomorrow:
https://docs.google.com/document/d/1rs13t8oTTyYgnm_0bGF_K38rRzG7WtYa0ztbD2K8T8E/edit#

If people are happy with it, I can turn it into proposal where we can vote on it.

bcantrill:
It remains unclear to me what problem is being solved here: if people wish to meet and discuss, CNCF doesn't prevent that -- but if some sort of implicit endorsement is being sought to be given by the TOC and/or CNCF, we should be much more clear about what is being done here and why.  Why not let an informal group form and then come to the TOC when they feel that the association with the CNCF would be helpful to them?

caniszczyk:
@bcantrill they are already discussing this and want a place to do it in a formal fashion as the end result may be a CNCF project proposal (these are CNCF members wanting to do this). We should provide it for them along with some simple rules (transparent meetings, short lived charter, etc)

The purpose is clearly laid out here, feel free to comment on the PR:
https://github.com/cncf/toc/blob/add-working-group-process/process/working_groups.adoc
https://github.com/cncf/toc/pull/6/files

caniszczyk:
This is finally done: https://github.com/cncf/toc#working-groups

PR #6, opened 2016-04-20
Add Working Groups proposal

caniszczyk:
Solidifies Working Group proposal process #3 

Signed-off-by: Chris Aniszczyk caniszczyk@gmail.com

### DIFF
--- a/process/working_groups.adoc
+++ b/process/working_groups.adoc
@@ -1,0 +1,56 @@
+== CNCF Working Groups v1.0
+=== Purpose
+The purpose of working groups are to study and report on a particular question and make recommendations based on its findings, along with providing an avenue for non-members to participate. The end result of a working group may be a new project proposal or a report detailing their findings. The intention of working groups is not to host a full project or specification. Working Groups can be formed at any time but must be ratified by the CNCF TOC. Once formed the work defined in the Working Group charter is the responsibility of the WG rather than the CNCF TOC. Working Groups are also expected to adhere to the CNCF IP Policy and Code of Conduct (https://cncf.io/governance).
+
+Working Groups are meant to be focused on a specific deliverable or question.  The charter, as approved by the TOC should be _crisp_ and focused.  There should also be an expected delivery date in the charter (possibly 3 to 6 months, possibly longer, but any Working Group that is envisioned to last longer than a year should be most carefully considered).
+
+The TOC can, with the agreement of the Working Group update the charter if necessary, but major shifts should trigger the creation of a new Working Group, rather than a retasking of an existing one.
+
+The Working Group, or the TOC, may dissolve a Working Group if they determine that the Working Group will be unable to meet the deliverable within the timeline defined in the charter.
+
+=== Process
+ * Starting a Working Group
+ * Working Group Governance
+
+=== Current Working Groups
+ * Service Broker: https://github.com/cncf/wg-servicebroker
+
+=== Starting a WG
+A Working Group is established by first defining a charter that can be ratified by the TOC. A charter is a statement of purpose, a list of responsibilities and a list of initial membership. A working group needs 3 initial members. These should be individuals already undertaking the work described in the charter. The list of responsibilities should be specific. Once established, these responsibilities are no longer governed by the TOC and therefore should not be broad or subjective. The only recourse the TOC has over the working group is to revoke the entire charter, take on the work previously done by the working group themselves or decide the work is no longer necessary.
+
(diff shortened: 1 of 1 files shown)


caniszczyk:
I'd like the @cncf/toc to take a final look at this and cast a vote. We've had some time to discuss and review this with the wider community and we are on our way to have our first WG proposed in Service Broker land: https://docs.google.com/document/d/1JnjndNwBB9mct91MLofzNrJwew_MJgwPXlULzrKx14Q/edit#heading=h.32tcpqhictr1

So RFC and +1 when you're ready

duglin:
The term "working group" implied to me a short-lived group of people who want to focus on a particular topic of discussion - usually to do some investigation and come back with some kind of recommendation for the TOC to consider. And I think, for the most part, this proposal aligns with that.  However, reading through it, when it talks about formal (heavy weight) concepts like charters that's when I get a little worried that we may be formalizing this a bit too much.

If the original intent is to have a mechanism by which the TOC can ask a group of people to "go off and think about xxx" then perhaps it would be better to keep it a more light-weight - something like:
- a TOC member(s) request some investigation be done
- a TOC member volunteers to lead the work. They solicit additional volunteers by sending a note to the mailing list. They decide how best to work/communicate among themselves. If needed we can help by establishing a localized mailing list. The list of contributors need not be limited to TOC or even CNCF members.
- the TOC leader periodically reports back on status to the TOC during the regular calls
- when completed a final report/recommendation (or whatever the requested outcome is) is presented to the TOC

In the end, this is really nothing more than a group of folks talking and that can happen even w/o us defining a process (obviously). The only thing this adds is the notion of the TOC specifically asking for, and knowing about, the work so it gives the TOC a chance to "see how things are going" occasionally.

So, in terms of a formal proposal, perhaps all we need is a small update to our charter listing the above bullets as the mechanism by which the TOC requests that a bit of investigation work be done on their behalf.

caniszczyk:
@duglin something simple could be proposed at the GB level to modify the charter (we already have something in place for the End User TAB to form Special Interest Groups, section 8)

In the end, this proposal makes it so folks outside the TOC can propose something and also a way to have non-members involved with the community, on top of other simple governance rules (2/3 super majority voting, open/transparent meetings). I believe having some minimal governance structure in place would be useful as some of the topics may be contentious.

I prefer that the TOC doesn't have to be involved unless they want to, WGs should be short lived and fairly autonomous outside of being ratified by the TOC. A charter is a simple paragraph stating what the WG hopes to accomplish.

In the end, it's important to write something down so we can agree to it. If you want to propose something shorter, I'm all for it, this is meant to be a v1.0 to get us started.

bgrant0607:
It seems to me that the main purpose is to provide a forum for interested parties to work together on a proposal, and to keep that forum/process open.

Once there is a proposal, it should be announced in a public enough place that interested parties outside the TOC can have an opportunity to provide input.

Given that, it seems something far less formal could be sufficient, and I don't think resources required would be significant. 

If we agree a proposal in a given area would be useful, we could agree to create a slack channel and mailing list (provided we get some reasonable mailing list solution), and request that meetings be announced in advanced on the relevant mailing list and that someone takes notes, which would then be sent to the relevant mailing list.. In principle, the channel and mailing list wouldn't remain active forever (though archives could), but I don't think we know yet what timeframe would be reasonable, so I wouldn't specify a specific time period at this point.

duglin:
I'm not sure yet on how people view working groups.  I see, at least, two ways to view them:

1 - something a bit formal that requires TOC approval. I see this as typically something focused on a particular question/problem that the TOC wants people to investigate. This would prevent a swarm of resources being setup (e.g. mailing lists) for activities that quickly die or people lose interest in - meaning wasted resources.  With a TOC sanctioned activity I would expect periodic status updates to the TOC which keeps pressure on them to keep moving forward, or formally shut down.

2 - something light-weight that anyone in the CNCF can initiate.  My general sense is that people might prefer this because its more open and doesn't require TOC's input or approval.  In this mode while a mailing list might be nice, even that could require some level of busy work for someone to setup a mailing list each time. 

This got me thinking about github issues.  I typically see issue discussions via email so they're sort of like mailing lists. They can be initiated by anyone. Anyone is free to join in.  They're easily started, ended/closed.  A history of the chats are saved. People are free to have conf calls if needed - and can announce and post summaries of them in the issue as a comment.  Do we really need anything more?  It may not sounds as fancy as a "working group", but I'm having a hard time seeing why we need anything more.

bgrant0607:
@dankohn @monadic @caniszczyk This proposal was never approved. This is a process issue we need to resolve, along with RFPs. 

Has the governance board undertaken formation of working groups under "establish and oversee any committees created to drive the mission of CNCF" in the charter?
https://cncf.io/about/charter

caniszczyk:
WGs have been formalized now: https://github.com/cncf/toc/blob/master/README.md#working-groups

PR #17, opened 2016-10-20
Create initial fluentd project proposal for RFC

caniszczyk:
http://www.fluentd.org/


caniszczyk:
@edsiper each repo being contributed has a respective AUTHORS file with the actual committers for the respective project?

PR #34, opened 2017-03-21
Update kubernetes proposal doc

rothgar:
I'm not sure if proposals should be updated or left in time as a snapshot of the request. If the former is true here is an updated link for containerized apps to bring people to the right page the first time instead of through 2 manual redirects and scrolling down the page.


caniszczyk:
@rothgar thanks for the small fix

PR #54, opened 2017-10-10
Plural "project" for 'CNFC is home for...'

chasemp:
It seems as if the sentence which leads with 'CNCF is home for...' '"project"' should be plural to match the usage slightly above 'And we put projects front and center._'.  Even though project is quoted indicating the varied usage of the term I think this is appropriate.

CNCF is a home for several kinds of “projects” where community collaboration furthers the goals of the CNCF community:
1. Open source software projects, e.g., Prometheus.


idvoretskyi:
LGTM

dankohn:
This is a different part of speech and the singular reads better to me.

PR #56, opened 2017-10-13
Fix broken link for Jaeger

spacewrangler:
404 caused by a type in the link to Jaeger repo. Instead of fixing that, changing the link to point to the project page instead of the repo.


caniszczyk:
Thanks, fixed it another way: https://github.com/cncf/toc/commit/17f71d37c5ce15d1eadc32c7213097012937e75a

PR #70, opened 2017-11-16
Project proposal: Istio Service Mesh

nilesh32:
As requested during the November 7th, 2017 CNCF TOC meeting, we submit Istio for consideration to be included in a CNCF project.

### DIFF
--- a/proposals/istio.adoc
+++ b/proposals/istio.adoc
@@ -1,0 +1,90 @@
+== istio
+
+*Name of project*: Istio +
+
+*Description*: +
+
+As monolithic applications are decomposed into microservices, software
+teams have to worry about the challenges inherent in integrating services
+in distributed systems: they must account for service discovery, load
+balancing, fault tolerance, end-to-end monitoring, dynamic routing for
+feature experimentation, and perhaps most important of all, compliance and
+security. +
+
+Istio eases the monolith to microservice transformation and the
+microservice application architecture by systemically injecting proxies
+into the network paths between services. The proxies form a service mesh
+that is dynamically configured and managed by the Istio control
+plane. Istio adds traffic management to microservices and creates a basis
+for value-add capabilities like security, monitoring, routing, connectivity
+management and policy. +
+
+Using the Istio control plane, developers can experiment with different
+features in production, deploy canary releases, and improve the reliability
+of their services without impacting the system as a whole. Operators gain
+fleet-wide visibility through extensive metrics and traces without having
+to rely on developers to emit metrics in their application code. They can
+improve end-to-end security, compliance, and the availability of services
+with minimal disruption to the development process. +
+
+The control plane is agnostic of the underlying orchestration platform
+(e.g., Kubernetes, Mesos, Cloud Foundry, Nomad, or even just VMs and bare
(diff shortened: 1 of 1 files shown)


nilesh32:
Signed-off-by: Nilesh Patel pateln@us.ibm.com

nilesh32:
Signed-off-by: Shriram Rajagopalan <shriram@us.ibm.com>
Signed-off-by: Nilesh Patel <pateln@us.ibm.com>

debianmaster:
+1

jessfraz:
In my opinion projects should have adoption before being added. This project is _so_ new. Why do you need to be in CNCF? Is it just for status? Foundations, imho, are not for marketing.

proposals/istio.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+*Sponsor / Advisor from TOC*: Brian Grant +
+*Preferred maturity level*: incubating +
+*Unique identifier*: Istio +
+*License*: Apache 2.0 +
+*Source control repositories*:
+
+  * https://github.com/istio/istio[https://github.com/istio/istio] +
+  * https://github.com/istio/api[https://github.com/istio/api] +
+  * https://github.com/istio/proxy[https://github.com/istio/proxy] +
```
sdake: with the monorepo merge, these repos are no longer accurate.

proposals/istio.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+*Contributor statistics*: 71 contributors (most activity from IBM/Google).
+2 orgs with commit access (IBM/Google). Other contributors include
+
+ * RedHat
+ * Tigera
+ * Yahoo!
+ * Apprenda
+ * Concur
+ * AT&T
```
sdake: not sure if it matters, however, the contributor list is much larger.

proposals/istio.adoc
```diff
@@ -0,0 +1,90 @@
...
+control plane thus creates a basis for platform independent value-add
+capabilities like security, monitoring, routing, connectivity management
+and policy enforcement, in addition to easing migration across cloud
+providers and orchestration platforms. +
+
+In summary, the service mesh weaved and managed by Istio provides a uniform
+platform-independent way to connect, manage, and secure microservices. +
+
+*Sponsor / Advisor from TOC*: Brian Grant +
+*Preferred maturity level*: incubating +
```
sdake: The incubation stage requires three real deployments of suitable scope:  Reference: https://github.com/cncf/toc/blob/master/process/graduation_criteria.adoc

Might consider the inception stage?  I suspect after 1.0 is released, incubation will be a nobrainer.

ddysher:
Thanks for the write up. We've been evaluating istio in v0.2 release, there're two major concerns for us to use istio in production.

- Control plane is single point of failure.  While envoy is battle tested, failure of istio pods can cause problem to the mesh
- Latency is a bit higher than expected, seems every request will be sent to control plane

This makes me wonder what is the scalability limit of istio?  I believe v0.3 has improvement around performance and stability (like request is cached in mixer), but I didn't find a published documentation around this.

monadic:
@jessfraz the primary reason projects have inception status (applicable for early stage) is to promote collaboration.  As you imply this could generate marketing, but is that so bad that inception should be blocked?

PR #98, opened 2018-03-15
Update README.md - projects status

lukaszgryglicki:
Update:
Kubernetes Incubating -> Graduated
Linkerd, Rook Inception -> Sandbox


caniszczyk:
@lukaszgryglicki thanks Lukas I have something queued already to update all of our docs/processes with the new names but have put it on hold to the next TOC meeting where the sandbox name may change

PR #119, opened 2018-05-30
README: Fix a link for telepresence

gaocegege:
I think the right link is https://github.com/telepresenceio

PTAL


caniszczyk:
correct, thank you!

PR #132, opened 2018-07-16
wording

Fashaun:
Fix wording in chinese


dankohn:
My understanding is that the current wording is correct for simplified Chinese.

PR #145, opened 2018-08-20
Add a requirement regarding security audits

caniszczyk:
A part of achieving a CII Badge involves in setting up a security disclosure process, which is a great practice for all open source projects to have. However, not all security disclosure processes are tested so the TOC is considering the requirement moving forward to have CNCF projects go through a third party security audit which helps test the security disclosure process.

Some examples here:
https://github.com/envoyproxy/envoy#security-audit
https://coredns.io/2018/03/15/cure53-security-assessment/

### DIFF
--- a/process/graduation_criteria.adoc
+++ b/process/graduation_criteria.adoc
@@ -23,7 +23,7 @@ To graduate from sandbox or incubating status, or for a new project to join as a
 
  * Have committers from at least two organizations.
  * Have achieved and maintained a Core Infrastructure Initiative https://bestpractices.coreinfrastructure.org/[Best Practices Badge].
- * Have completed an independent and third party security audit with results published (i.e., https://github.com/envoyproxy/envoy#security-audit)
+ * Have completed an independent and third party security audit with results published of similar scope and quality as the following example (including critical vulnerabilities addressed): https://github.com/envoyproxy/envoy#security-audit
  * Adopt the CNCF https://github.com/cncf/foundation/blob/master/code-of-conduct.md[Code of Conduct].
  * Explicitly define a project governance and committer process. This preferably is laid out in a GOVERNANCE.md file and references an OWNERS.md file showing the current and emeritus committers.
  * Have a public list of project adopters for at least the primary repo (e.g., ADOPTERS.md or logos on the project website).
--- a/process/graduation_criteria.adoc
+++ b/process/graduation_criteria.adoc
@@ -23,7 +23,7 @@ To graduate from sandbox or incubating status, or for a new project to join as a
 
  * Have committers from at least two organizations.
  * Have achieved and maintained a Core Infrastructure Initiative https://bestpractices.coreinfrastructure.org/[Best Practices Badge].
- * Have completed an independent and third party security audit with results published of similar scope and quality as the following example (including critical vulnerabilities addressed): https://github.com/envoyproxy/envoy#security-audit
+ * Have completed an independent and third party security audit with results published of similar scope and quality as the following example (including critical vulnerabilities addressed): https://github.com/envoyproxy/envoy#security-audit and all critical vulnerabilities need to be addressed before graduation.
  * Adopt the CNCF https://github.com/cncf/foundation/blob/master/code-of-conduct.md[Code of Conduct].
  * Explicitly define a project governance and committer process. This preferably is laid out in a GOVERNANCE.md file and references an OWNERS.md file showing the current and emeritus committers.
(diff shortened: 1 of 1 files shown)


yurishkuro:
A couple of questions
  * what is the cost of 3rd party audit?
  * how much value is in a (one time) security audit for projects with frequent releases?

caniszczyk:
@yurishkuro CNCF would cover the costs of any audit, we have done so already for a few projects who were part of a pilot program

caniszczyk:
@yurishkuro regarding the value, it's hard to say, it definitely ensures that the security disclosure process works at one point in time. It can also help a project when adopted by large enterprises who have selection criteria that involve having an audit, for example, this was the case with Envoy recently: https://envoyproxy.slack.com/archives/C78M4KW76/p1534181685000447

process/graduation_criteria.adoc
```diff
@@ -23,7 +23,9 @@ To graduate from sandbox or incubating status, or for a new project to join as a
 
  * Have committers from at least two organizations.
  * Have achieved and maintained a Core Infrastructure Initiative https://bestpractices.coreinfrastructure.org/[Best Practices Badge].
+ * Have completed an independent and third party security audit with results published (i.e., https://github.com/envoyproxy/envoy#security-audit)
```
quinton-hoole: nit: e.g, not i.e.
We will need to write down in a bit more detail (elsewhere) what the requirements of the audit are.  To start with, something like "of similar scope and quality as the following examples" might suffice.

quinton-hoole: Also, lets be clear whether the critical vulnerabilities identified in such an audit need to be addressed before graduation.

quinton-hoole: And whether or not such an audit needs to be performed periodically, e.g. annually.

caniszczyk: thanks, tried to address your concerns with an updated sentence

JustinCappos:
@yurishkuro
I'll also mention that good security audits cover much more than the
existing code base.  Auditing the security architecture, design, tradeoffs,
etc. is also of great value across even completely different
implementations of the same project.

bgrant0607:
In addition to the audit and existing CII standards, we probably should look at security and quality more holistically.

The CII criteria are a good start, but don't cover everything:
https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md

For instance, I don't think code review is even recommended.

Other possible areas include test coverage, design documentation, security team and response procedures, dependency analysis automation, and observability (at least for server components). Kubernetes falls short in a number of those areas, so we are working on defining a higher bar.

david-a-wheeler:
@bgrant0607 - thanks for referring to the CII Best Practices passing badge criteria (which [Kubernetes already meets](https://bestpractices.coreinfrastructure.org/en/projects/569))!

However, https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md is only the criteria for the *passing* badge.  We have two higher level badges, silver and gold, which might in part be what you're looking for. [The silver and gold criteria are also posted on GitHub](https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/other.md).  Each level requires that you meet the previous one (e.g., to get silver, you first have to get a passing badge).  Please take a look!  If you have questions, feel free to ask, I'd love to help.

mayakacz:
This is a very reasonable minimum bar.

process/graduation_criteria.adoc
quinton-hoole-2: We seem to still be missing wording around whether critical vulnerabilities need to be addressed before graduation, and whether periodic (or per release) audits are required to remain in "graduated" state.  Given that the wording is missing, I assume that the implied answer to both of the above is "no", or "at the discretion of the TOC".  I would strongly encourage making that explicit, to avoid confusion.

mattklein123: I agree with @quinton-hoole-2 that we should make these explicit. IMO requiring that critical vulns be fixed prior to graduation is a given. Personally, I think that we should at least "highly encourage" periodic security audits, although I recognize funding needs to be considered.

mattklein123:
+1 from me on this as an independent change, although I would still like to discuss the periodic aspect separately if folks are more comfortable doing it that way.

brendandburns:
This seems reasonable to me as a place to start.

caniszczyk:
No objections, will merge it in. We will deal with the periodic aspect later as that's more complicated given the recurring budget requirements + balancing security audit contractor availability.

Thanks all!

PR #151, opened 2018-09-16
Add a way to recognize previous TOC members

caniszczyk:
We should list previous TOC members and thank the for their service.


quinton-hoole:
lgtm

caniszczyk:
no strong objects from anyone, merging this in after 10 days, thanks all

PR #159, opened 2018-10-02
Updating zoom link for TOC call

mattfarina:


caniszczyk:
updated already, thanks @mattfarina

PR #200, opened 2019-03-02
Add Zhipeng as new ToC Contributor

hannibalhuang:
As co-lead of the Kubernetes Policy WG and in-forming SAFE WG, I would like to be a ToC Contributor serving as a liaison between CNCF ToC and these two groups on matters related to policy


quinton-hoole:
LGTM

Issue #214, opened 2019-03-22
RFC: redesign this repo's README and documentation


mattklein123:
I don't think the README for this repo, as a landing page for understanding the TOC, is in the best interest of our community. A few high level notes of things I would like to discuss doing:
- Leave overview, members, and mailing list / meeting / contact info
- Move all other sections into sub-pages, linked to from the primary README
- On the projects page, either remove it and link to the main CNCF site, or clearly break down between graduated, incubating, and sandbox.
- On the main page, have a section on "proposing a new project for the CNCF," with a brief overview and then further links into the contributing guidelines. The goal here should be to make contributors and the community feel very comfortable with the process without having to click through various un-linked pages to try to piece things together.
- On the main page, have a section on SIGs with a brief overview and links for further information, how to join, etc.

Is there anything else anyone would like to see?

xiang90:
> On the projects page, either remove it and link to the main CNCF site, or clearly break down between graduated, incubating, and sandbox.

I suggest we remove it from the toc repo and just link to main CNCF site.

There is a website guideline section. I am not sure how it relates to the toc.

lizrice:
Great idea, thanks for kicking this off @mattklein123. +1 all your suggestions. In addition

For meeting info, let's make it clear who can attend.  We can move the links to minutes & presentations on a separate page, in reverse chronological order.

As part of clarifying the whole new project process (in fact the project lifecycle), let's clarify the role of TOC sponsors.

lizrice:
Also need to document the SIG liaisons

xiang90:
@mattklein123 

Shall we start to execute the plan you proposed? I can help on creating PRs to finish it.

mattklein123:
@xiang90 if you have cycles to start working on it that sounds awesome. @lizrice WDYT?

lizrice:
Sounds great to me! I'll try to pitch in on some of this too

ultrasaurus:
Here's unmerged PR that has some suggestions along these lines for SIG content: https://github.com/quinton-hoole/toc/pull/3

PR #225, opened 2019-04-12
Update about community / project presentations

lizrice:
The link to scheduled community presentations points to a page that doesn't seem to have been maintained recently, so I'm proposing here that we archive it. Also clarify that project presentation requests are tracked on the backlog.

Falls under #214


caniszczyk:
Thanks @lizrice

PR #226, opened 2019-04-12
Add possibility of SIG assessment

lizrice:
Modified Project Acceptance Process to include the possibility of a SIG assessment step. What this assessment consists of is left open - could be tightened up later. 

Also noted that we generally want to know what companies are represented in the project committers.

* TOC folks, does this agree with your understanding of the process? 
* Is this significant enough to warrant a semver increment in the header? (Do people use / need the semver for processes like this?)

### DIFF
--- a/process/project_proposals.adoc
+++ b/process/project_proposals.adoc
@@ -23,6 +23,6 @@
 
 . *Project Acceptance Process*.
  .. Project is proposed via a GitHub issue
- .. TOC may request that a [CNCF SIG](https://github.com/cncf/toc/blob/master/sigs/cncf-sigs.md) assesses the project 
+ .. TOC may request that a https://github.com/cncf/toc/blob/master/sigs/cncf-sigs.md[CNCF SIG] assesses the project 
  .. Projects are required to present their proposal at a TOC meeting
  .. Projects get accepted via a 2/3 supermajority vote of the TOC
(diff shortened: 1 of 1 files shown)


caniszczyk:
+1

RFC @cncf/toc

process/project_proposals.adoc
```diff
@@ -22,5 +22,7 @@
  .. project logo in svg format (see https://github.com/cncf/artwork#cncf-related-logos-and-artwork for guidelines)
 
 . *Project Acceptance Process*.
+ .. Project is proposed via a GitHub issue
+ .. TOC may request that a [CNCF SIG](https://github.com/cncf/toc/blob/master/sigs/cncf-sigs.md) assesses the project
```
xiang90: @lizrice this is not a markdown file. so the link format does not apply. we need to change it to link[content]. see line 18 for example.

xiang90:
@lizrice 

+1. 

A side note, the item v on section 1 and item 4 on section 2 conflicts. If projects choose sandbox maturity level, we do not require 2/3 supermajority vote of the TOC. /cc @caniszczyk

ultrasaurus:
+1 :)

process/project_proposals.adoc
lizrice: thanks @xiang90, fixed

lizrice:
> A side note, the item v on section 1 and item 4 on section 2 conflicts. If projects choose sandbox maturity level, we do not require 2/3 supermajority vote of the TOC. /cc @caniszczyk

Raised separately as #274 so we can merge this PR

PR #227, opened 2019-04-13
Make it clear that sandbox projects must adopt CoC

caniszczyk:
per @lizrice

https://github.com/cncf/foundation/issues/36

### DIFF
--- a/process/graduation_criteria.adoc
+++ b/process/graduation_criteria.adoc
@@ -1,4 +1,4 @@
-== CNCF Graduation Criteria v1.2
+== CNCF Graduation Criteria v1.3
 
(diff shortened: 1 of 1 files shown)


process/graduation_criteria.adoc
```diff
@@ -7,6 +7,8 @@ Projects of all maturities have access to all resources listed at https://cncf.i
 === Sandbox Stage
 
 To be accepted in the sandbox a project must have at least 2 TOC sponsors. See the https://github.com/cncf/toc/blob/master/process/sandbox.md[CNCF Sandbox Guidelines v1.0] for the detailed process.
+ 
+ * Adopt the CNCF https://github.com/cncf/foundation/blob/master/code-of-conduct.md[Code of Conduct].
```
xiang90: This should be added after "must have at least 2 TOC sponsors" before "See the ... for detailed process" of line 9. We probably should just break down line 9 like Incubation Stage and Graduation Stage doc style.

PR #239, opened 2019-05-20
create readme for readability/maintainability

ultrasaurus:
small refactor to make this more approachable for newcomers
- provide a index to SIGs and high level description
- factor out proposed SIGs into their own doc 
- small change to front matter of process doc since it is no longer a proposal

The idea is that when a new SIG is created, it can be removed from proposed.md and added into the README index, then it is easy to see the list of SIGs to be created and for people to find SIGs that are available to join


caniszczyk:
Thanks @ultrasaurus

PR #240, opened 2019-05-20
Correct spelling errors of 'Sandbox Exit Requirements'

zhangthen:
The 2/3 of the votes are larger than one, and the 'vote' should be plural.


lizrice:
Strictly I think this should be "a 2/3 majority vote"

PR #244, opened 2019-05-21
Spanish translation

alombarte:
Translated it because it was a short text :)


caniszczyk:
https://github.com/cncf/toc/blob/master/DEFINITION.md#espa%C3%B1ol

PR #245, opened 2019-05-28
Add link to CNCF IP policy

lizrice:


caniszczyk:
SGTM, this is implied by the charter anyways so I will add it w/o a formal vote

PR #247, opened 2019-05-30
Update DEFINITION.md translate into Spanish

zdeisidr:
Update DEFINITION.md translate into Spanish (Spain)

### DIFF
--- a/DEFINITION.md
+++ b/DEFINITION.md
@@ -2,7 +2,7 @@
 
 *Approved by TOC: 2018-06-11*
 
-[中文版本](#中文版本) | [日本語版](#日本語版) | [한국어](#한국어) | [Português Brasileiro](#português-brasileiro) (in Chinese, Japanese, Korean and Brazilian Portuguese below)
+[中文版本](#中文版本) | [日本語版](#日本語版) | [한국어](#한국어) | [Português Brasileiro](#português-brasileiro) | [Español de España](#español-españa) (in Chinese, Japanese, Korean, Brazilian Portuguese and Español de España below)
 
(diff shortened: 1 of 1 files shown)


dankohn:
Could you please get some other native Spanish speakers to review, preferably from other companies. Also, we'd prefer to do a Spanish that is equally understandable in Spain and Latin America, if that's feasible.

alombarte:
Hi, I did a lot of translations in the past as an open source contributor, I am not a professional but I did it for fun quite a few times. I am a native speaker of both Spanish and Catalan.

**There is duplication with the issue I opened a while ago #244**

Anyway, I reviewed the current PR and found that needs some corrections. Some examples are:

> permiten a las organizaciones a construir

permiten a las organizaciones construir ( remove extra "a")

>...y APIs declarativas ejemplifican este alcance

alcance means "reach". The original text uses "approach" instead. So 
"y las API declarativas son un ejemplo de esta aproximación."

> Éstas técnicas habilitan...

"Estas" doesn't have an accent

> resilentes, manejables y observables.

resilientes (missing an "i")

> redes de servicio

This would mean "service networks" and not "service mesh". In my translation, I didn't do any adaptation as there no consensus on the translation yet. If I had to I would go for something like "malla de servicio", but quite weird.

Finally I wouldn't choose "Español de España" and would solely write Spanish.

The text is perfectly understandable though correcting the grammar. I am not going to say I prefer my own translation ¯\_(ツ)_/¯.

caniszczyk:
https://github.com/cncf/toc/blob/master/DEFINITION.md#espa%C3%B1ol

PR #258, opened 2019-06-20
Explain abbreviation "GB"

lfrancke:
The abbreviation "GB" is used later but never explained. This PR is super simple and all it does is it adds the abbreviation to the full explanation.


caniszczyk:
LGTM thanks

PR #267, opened 2019-07-18
Add comparison and end user info to proposals

lizrice:
Per TOC meeting 16-July-2019, proposals (at any maturity level) should include a comparison with similar projects, and information about who is using the project

### DIFF
--- a/process/project_proposals.adoc
+++ b/process/project_proposals.adoc
@@ -6,6 +6,7 @@
  .. name of project (must be unique within CNCF)
  .. project description (what it does, why it is valuable, origin and history)
  .. statement on alignment with CNCF charter mission
+ .. comparison with similar projects (inside or outside the CNCF), including what differentiates this project
  .. sponsor from TOC (sponsor helps mentor projects)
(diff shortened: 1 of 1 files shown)


caniszczyk:
Thanks @lizrice

process/project_proposals.adoc
```diff
@@ -19,6 +20,7 @@
  .. release methodology and mechanics
  .. social media accounts
  .. community size and any existing sponsorship
+ .. who is currently known to be using the project? Are they using it in production and at what scale? (It may be hard to obtain accurate data for this, but any supporting evidence of usage is helpful)
```
quinton-hoole: Note that one of the primary motivations behind Sandbox was to provide a neutral home to facilitate early-stage multi-company collaboration efforts, even for cases where those efforts have not yet reached a stage of being used. See https://www.cncf.io/sandbox-projects/ for more detail.

How do we propose to accommodate this goal? I assume that "signs of being used" would be considered "good to have" but not a blocker for entry into the Sandbox?

lizrice:
@quinton-hoole yes, exactly that. We realised we're asking the question "who is using this" at all stages anyway, so it makes sense to document it here

PR #291, opened 2019-09-12
Adding Polish translation for CNCF Definition

mtjakobczyk:
Adding Polish translation for CNCF Definition

### DIFF
--- a/DEFINITION.md
+++ b/DEFINITION.md
@@ -67,7 +67,7 @@ diese Innovationen für alle zugänglich zu machen.
 
 ## Polski
 
-Technologie **Cloud Native** pozwalają organizacjom tworzyć i wykorzystywać skalowalne aplikacje w nowoczesnych, dynamicznych środowiskach takich jak chmury publiczne, prywatne czy hybrydowe. Przykładami tego podejścia technologicznego są kontenery aplikacyjne (ang. *containers*), mikrousługi (ang. *microservices*), platformy dla mikrousług (ang. *service mesh*), niezmienna infrastruktura (ang. *immutable infrastructure*) i deklaratywne interfejsy aplikacyjne (ang. *declarative APIs*).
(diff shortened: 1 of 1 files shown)


DEFINITION.md
```diff
@@ -65,6 +65,14 @@ Die zugrundeliegenden Techniken ermöglichen die Umsetzung von entkoppelten Syst
 Die Cloud Native Computing Foundation fördert die Akzeptanz dieser Paradigmen durch die Ausgestaltung eines Open Source Ökosystems aus herstellerneutralen Projekten. Wir demokratisieren modernste und innovative Softwareentwicklungs-Patterns, um
 diese Innovationen für alle zugänglich zu machen.
 
+## Polski
+
+Technologie **Cloud Native** pozwalają organizacjom tworzyć i wykorzystywać skalowalne aplikacje w nowoczesnych, dynamicznych środowiskach takich jak chmury publiczne, prywatne czy hybrydowe. Przykładami tego podejścia technologicznego są kontenery aplikacyjne (ang. *containers*), mikrousługi (ang. *microservices*), platformy dla mikrousług (ang. *service mesh*), niezmienna infrastruktura (ang. *immutable infrastructure*) i deklaratywne interfejsy aplikacyjne (ang. *declarative APIs*).
```
derberg: Hey, small comment from my side regarding `kontenery aplikacyjne`
In Polish, it sounds awkward (like most of polish versions of English tech words btw :)). Shouldn't it be just `kontenery` or at least `wirtualne kontenery` https://pl.wikipedia.org/wiki/Docker_(oprogramowanie). I'm afraid `applikacja` (application) is to narrow in context of CNCF definition. `kontenery (ang. *containers*)` should be enough imho.

you can of course ignore me, just sharing my feedback. I have to admit I have no experience in translating English tech to Polish tech and I do not read tech articles/books in Polish

mtjakobczyk: Good spot. Thank you.

I agree with your comment. I will change it to `wirtualne kontenery`. I believe `kontenery` alone would be still too vague, even though the term `kontener` is nowadays rather obvious to us.

Indeed, it is really challenging to translate some technical words. On the other hand we could take this as an opportunity to agree on a common translation. I remember my time at Warsaw University of Technology and, to some degree philosophical, discussions whether to mix Polish and English or strive to find Polish terms for everything. I believe we should go with the second approach when dealing with **definitions**.

derberg: as you noticed, these discussions are usually "philosophical" and sometimes even based on way to "patriotic" feelings about the language. I always say `mikroserwisy` and `serwis mesz` and hate when someone talks to me about `zwinne programowanie` (agile) and I literally have no idea what they say and waste my brain for on-the-fly translation that hey ment `agile` (`edżajl`) 😄 

but yeah, for me this is a subject for a discussion supported by at least one 🍺

caniszczyk:
dziękuję!

PR #292, opened 2019-09-13
Kyma proposal for CNCF Sandbox

derberg:
This is a proposal for making Kyma a part of the CNCF Sandbox.
As mentioned in the application form, we did not discuss upfront the project with any TOC members to get sponsors upfront. Our understanding is that first, we create this PR and then get scheduled for TOC meeting and present Kyma first.

Please share your feedback and let us know if anything is unclear or missing

### DIFF
--- a/proposals/kyma.md
+++ b/proposals/kyma.md
@@ -4,7 +4,13 @@
 
 *Project Description:*
 
+Kyma goal is to solve 2 problems:
+- Difficulty of moving monolith legacy software into cloud
+- Challange related to choosing proper cloud-native tools to build production ready solutions on top of Kubernetes
+
 https://kubernetes.io/blog/2019/05/23/kyma-extend-and-build-on-kubernetes-with-ease/
+
+
 
 *Statement on alignment with CNCF mission:*
 
--- a/proposals/kyma.md
+++ b/proposals/kyma.md
@@ -6,7 +6,7 @@
 
 Kyma goal is to solve 2 problems:
 - Difficulty of moving monolith legacy software into cloud. Enable legacy products to be enhanced and evolved with a side-by-side flexible cloud native runtime, supporting both lightweight functions, static content or heavyweight microservices.
-- Challange related to choosing proper cloud-native tools to build production ready solutions on top of Kubernetes
+- Challenge related to choosing proper cloud-native tools to build production ready solutions on top of Kubernetes
 
 More details:
 Blog about Kyma on Kubernetes blog: https://kubernetes.io/blog/2019/05/23/kyma-extend-and-build-on-kubernetes-with-ease/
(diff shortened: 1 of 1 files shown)


caniszczyk:
RFC @cncf/toc 

@amye to schedule

iancrowther:
:-) nice

sebgoa:
Parsing quickly through this proposal I am not seeing kubeless being mentioned as a dependency, even though there is a lot of kubeless reference in the codebase. It would be nice to clarify kubeless usage.

It is also interesting to note the dependency on two big codebases Istio and Knative which are themselves not part of CNCF.

bgrant0607:
@sebgoa Just on the dependency issue: Kubernetes depended on Etcd and Docker at the time it was donated to CNCF, and now has lots of dependencies on other open-source repos. I'm sure our other projects have dependencies also, not all of which are in CNCF. Dependencies aren't inherently disqualifying.

derberg:
@sebgoa Hey, thanks for taking the time to look into the proposal! much appreciated. 

We mention Kubeless here:
- https://github.com/cncf/toc/pull/292/files#diff-af8780e43e381d15c475e3b1b3c536c7R19
- https://github.com/cncf/toc/pull/292/files#diff-af8780e43e381d15c475e3b1b3c536c7R188

We are in the process of migrating to Knative as mentioned in the PR. More details: the experimental implementation behind the feature flag is already in Kyma https://github.com/kyma-project/kyma/tree/master/resources/function-controller. The plan is to not use Kubeless at all.

---

Regarding dependency to Istio and Knative. Yes, we depend on those projects that are in CNCF Landscape.
- We also work now on K8s controllers with Ory on https://github.com/ory/oathkeeper-maester and https://github.com/ory/hydra-maester to make sure in Kyma you also have a way to enable OAuth2 using CRD. Ory is also on [CNCF Landscape](https://landscape.cncf.io/). 
- We also use Dex, Minio, Loki, Grafana
- W depend on custom components we had to write on our own as they were not available in the community and we did not add them to this PR because they can be used without Kyma too, like for example: https://github.com/kyma-project/helm-broker.

Not all dependencies we need are in CNCF sandbox/incubator/graduate. Is this a problem? or you just want us to make it more transparent in this PR. I can extend the PR more, not a problem for me.

I would only need guidance on the level of detail you would expect. Thing is that for example, Kyma Console UI is heavily dependent on React and still Angular (migration phase to React) and https://github.com/SAP/luigi microfrontend framework. Is this needed in this PR? 
I would appreciate some advice here.

sebgoa:
My 2 cts is that you should really remove kubeless before anything happens, since this project (which I started) is now maintained by a single developer and its future is quite uncertain.

On the dependency issue, I understand it. It just seems strange that the control plane for a CNCF project would be non-CNCF projects that run within a CNCF project. I am not asking for answers but only making an observation.

derberg:
@sebgoa thank you. Yesterday we had our community public call where one of the topics we handled was Kubeless vs new Function controller that we are working on, so the latest status is shared https://youtu.be/9tJOn6qRR2U?t=881 tl;dr move to Knative is in progress

jbeda:
We just discussed this in our face to face TOC meeting and don't think that Kyma is a good fit for the CNCF.  

Specifically, the [TOC principles](https://github.com/cncf/toc/blob/master/PRINCIPLES.md#no-kingmakers--one-size-does-not-fit-all) states:

> No one stack. The TOC does not pick a “winning stack” - i.e., vertically integrated set of projects as a solution for multiple application problems. Instead by encouraging interop, we hope that a range of patterns & “stacks” will emerge."

There might be an appropriate place elsewhere in the LF for the project.  Reach out to @caniszczyk if you want to start exploring those options.

PR #295, opened 2019-09-18
Update to match updated charter

caniszczyk:


caniszczyk:
The GB has approved a charter change.

PR #325, opened 2019-12-16
Update projects.md

amye:
Adding audits + announcements for projects that have them


caniszczyk:
There's a lot more here @amye but thanks for the great start!

PR #327, opened 2019-12-19
Create process for Sandbox annual review

lizrice:
We have existing [documentation](https://github.com/cncf/toc/blob/master/process/sandbox.md) that says that Sandbox projects should have an annual review. This PR creates a process for that annual review. 

(This PR supercedes the [Google doc](https://docs.google.com/document/d/1OwahJ8aQ880_JA6H6uLwsaQZ9sKWipcK_4RQ1SDfnGc) originally circulated for initial comments.)

### DIFF
--- a/process/sandbox-annual-review.md
+++ b/process/sandbox-annual-review.md
@@ -1,0 +1,43 @@
+# Sandbox annual review 
+
+Sandbox projects are subject to an annual review by the TOC. This is intended to be a lightweight process to ensure that projects are on track, and getting the support they need.
+
+## How to file your annual review
+
+CNCF staff will notify the project maintainers when the project review is due. 
+
+Project maintainers are responsible for agreeing between them who will complete the annual review. One of the maintainers should create the review in GitHub under [cncf/toc/reviews](https://github.com/cncf/toc/tree/master/reviews). 
+
+* Raise a PR titled *[Project name] Annual Review*
+* The PR should include a file called `annual-<project name>-<year>.md` (for example, `annual-amazingproj-2019.md`) with the contents described below
+* Send an email to the TOC mailing list so that the community knows the PR is there and can comment on it
+
+If your annual review isn’t complete within two months of notification, we will take this as a sign that the project is not under active maintenance and the TOC is likely to decide to archive the project. 
+
+## Review outcomes
+
+The outcome of the annual review is either:
+
+* At least three (two before the TOC grows to 11 members) TOC members agree to continue to sponsor the project in Sandbox, or
+* If enough TOC members don’t agree to continue to sponsor the project, we will discuss archiving or other next steps with you. 
+
+Additionally, the TOC might recommend that you apply for Incubation stage. This requires extra work and due diligence so it’s not a possible outcome directly from this lightweight annual review. 
+
+It is fine for projects to stay in the Sandbox indefinitely while it is still active, but if a project has genuinely stalled we can save everyone’s effort by [archiving](https://github.com/cncf/toc/blob/master/process/archiving.md) it. 
+
+## Annual review contents
+
+Your annual review should answer the following questions: 
+
+* Include a link to your project’s devstats page. We will be looking for signs of consistent or increasing contribution activity. Please feel free to add commentary to add colour to the numbers and graphs we will see on devstats.
+* How many maintainers do you have, and which organisations are they from? (Feel free to link to an existing MAINTAINERS file if appropriate.)
+* What do you know about adoption, and how has this changed since your last review / since you joined Sandbox? If you can list companies that are end users of your project, please do so. (Feel free to link to an existing ADOPTERS file if appropriate.)
+* How has the project performed against its goals since the last review?
+* What are the current goals of the project? For example, are you working on major new features? Or are you concentrating on adoption or documentation? 
+* How can the CNCF help you achieve your upcoming goals? 
+* Do you think that your project meets the [criteria for incubation](https://github.com/cncf/toc/blob/master/process/graduation_criteria.adoc#incubating-stage)? 
+
+
+
+
+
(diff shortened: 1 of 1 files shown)


process/sandbox-annual-review.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Annual review contents
+
+Your annual review should answer the following questions: 
+
+* Include a link to your project’s devstats page. We will be looking for signs of consistent or increasing contribution activity. Please feel free to add commentary to add colour to the numbers and graphs we will see on devstats.
+* How many maintainers do you have, and which organisations are they from? (Feel free to link to an existing MAINTAINERS file if appropriate.)
+* What do you know about adoption, and how has this changed since your last review / since you joined Sandbox? If you can list companies that are end users of your project, please do so. (Feel free to link to an existing ADOPTERS file if appropriate.)
+* How has the project performed against its goals since the last review?
+* What are the current goals of the project? For example, are you working on major new features? Or are you concentrating on adoption or documentation?
```
fntlnz: Here I would emphasize that the defined goals are subject of next year's annual review where applicable.

process/sandbox-annual-review.md
```diff
@@ -0,0 +1,43 @@
...
+
+Sandbox projects are subject to an annual review by the TOC. This is intended to be a lightweight process to ensure that projects are on track, and getting the support they need.
+
+## How to file your annual review
+
+CNCF staff will notify the project maintainers when the project review is due. 
+
+Project maintainers are responsible for agreeing between them who will complete the annual review. One of the maintainers should create the review in GitHub under [cncf/toc/reviews](https://github.com/cncf/toc/tree/master/reviews). 
+
+* Raise a PR titled *[Project name] Annual Review*
```
fntlnz: Maybe include the year here too? Like

```
*[Project name] 2020 Annual Review*
```

process/sandbox-annual-review.md
```diff
@@ -0,0 +1,43 @@
...
+* The PR should include a file called `annual-<project name>-<year>.md` (for example, `annual-amazingproj-2019.md`) with the contents described below
```
fntlnz: I would recommend doing a file name like:

```
2019-amazingproj-annual.md
```

In this file order is significant to the current round of reviews for the current year. Let's say that you want to open all of them in an editor

```
code 2019*.md
```

process/sandbox-annual-review.md
```diff
@@ -0,0 +1,43 @@
...
+* Send an email to the TOC mailing list so that the community knows the PR is there and can comment on it
```
fntlnz: I would suggest to add a Link to the TOC mailing list here

process/sandbox-annual-review.md
```diff
@@ -0,0 +1,43 @@
...
+
+If your annual review isn’t complete within two months of notification, we will take this as a sign that the project is not under active maintenance and the TOC is likely to decide to archive the project.
```
kmova: should this be changed to "If your annual review is not submitted within two months of notification"? What if the project has submitted for annual review before or after the notification is received, but the TOC or the CNCF SIG groups couldn't pick it up in time to do the review due to their backlog?

process/sandbox-annual-review.md
```diff
@@ -0,0 +1,43 @@
+# Sandbox annual review 
+
+Sandbox projects are subject to an annual review by the TOC. This is intended to be a lightweight process to ensure that projects are on track, and getting the support they need.
```
kmova: If there are SIGs in place, will the annual process first go through the SIG or will it directly be submitted to TOC?

lizrice: I think this should be a single-pase process to keep it simple and lightweight. Will clarify.

process/sandbox-annual-review.md
lizrice: Seems a bit moot in that I would hope only one is ever open at once for a project! But sure, we can add it.

process/sandbox-annual-review.md
lizrice: Yes, good idea. Would be even better if you used the suggest +- box to supply the link, to make it easy to act on :-)

process/sandbox-annual-review.md
lizrice: Very good point, yes

process/sandbox-annual-review.md
lizrice: Well, no, that’s not strictly true. We’re not going to penalize a project if they didn’t meet their previous year’s goals because better ones came along. It’s more about understanding whether they *have* goals and plans, and are actively working on them. I could clarify that though.

process/sandbox-annual-review.md
```diff
@@ -0,0 +1,45 @@
...
+
+To keep it lightweight, it is a single phase process - the review doesn't first pass through SIGs and then the TOC. The review document is shared with the community and SIG members can comment on each review, as can any other TOC contributor. The intention is to quickly reach a position where TOC members agree to continue sponsorship. 
+
+## How to file your annual review
+
+CNCF staff will notify the project maintainers when the project review is due. 
+
```
kmova: It's implicit that projects nearing 12 months since being submitted for sandbox or previous review will be notified. Also projects that are currently in review for incubation will be exempt from annual review. Should a note be added to state the criteria explicitly.

lizrice: The title covers the fact that it’s for Sandbox projects, and that it’s something that should happen annually. The reality is that right now we are well over one year for some projects, and because we have had no process, no notifications have happened yet, so there’s no point being too prescriptive right now over something we can’t immediately fix. It will be CNCF staff’s responsibility to do this in a timely fashion but there needs to be some leeway so everyone can catch up with the backlog. 

I disagree that projects currently in review for incubation should necessarily be exempt from annual review - I think this a matter for discretion for the TOC, CNCF staff and the project to agree amongst themselves (e.g. a project receiving their annual review notification might point at their incubation proposal as a response, and we can take it from there).

Issue #249, opened 2019-05-31
revisit the idea of needing 3 TOC Sponsors for Sandbox


caniszczyk:

lizrice:
TOC had some discussion on this and didn’t reach consensus, on the basis that it’s unclear what problem we are trying to solve and therefore whether three sponsors would solve that problem

ultrasaurus:
would be good to address open question about the responsibility of sponsors here: https://github.com/cncf/toc/issues/195

amye:
Closing as this changed when the TOC increased to 11.

PR #341, opened 2020-01-28
Merge when TOC size increases to 11

lizrice:
Update Sandbox sponsor count to reflect vote (https://lists.cncf.io/g/cncf-toc/message/4092)


lizrice:
Fixes #249

PR #353, opened 2020-02-07
Update README.md

raravena80:
- SIG-Runtime was approved/created
- SIG-ContribExp is being proposed (the name is in flux)


amye:
As of Tuesday, I think the name is more like Contributor Strategy.

raravena80:
Thanks @amye. Updated.

PR #321, opened 2019-11-20
Project on-boarding Flowchart

ultrasaurus:
This PR has two commits
1) move [original flowchart](https://docs.google.com/drawings/d/1Nvq188i3LGoWgtu1Jdk0wmpOcB6ZxvAc8_ppPygbnus/edit) into mermaid markdown
2) alternate flowchart to match observed or what seems to be the emerging process

Not wed to any of these ideas, just attempting to provide a little structure to clarify

image

Issue #344, opened 2020-01-30
Sandbox Project Proposal Template 


erinaboyd:
# CNCF SIG Sandbox Project Request Template & Promotion Template

# <Project> 

## <Promotion/Acceptance> to CNCF <Incubation/Sandbox>

**Primary aim: This template helps a project to prepare for a presentation to the SIG, and then helps the SIG share the collated information with the TOC.**

**Authors:**

# Background

**_Link to TOC PR_**

**_Link to Presentation_**

**_Link to GitHub project_**

## Future Plans

# Project Scope

## Clear project definition

**_Does the project have a clear and well defined scope_**

## Value-add to the CNCF ecosystem

**_Does the project have a clear value add to the current project ecosystem. How does it relate to other projects with overlapping capabilities._**

## Alignment with other CNCF projects

**_Does the project align and actively collaborate with other CNCF projects?_**

**_Does the project require any specific versions of projects (or APIs) to interoperate?  (e.g. K8s API, CSI, CNI, CRI)?_**

**_Does the project augment or benefit other CNCF projects?_**

## Anticipated use cases

**_What problems does the project solve?_**

## Alignment with SIG Reference Model

**_Does the project align with the SIG CNCF reference model and which capabilities does it require/provide at each level of the reference model._**

## High level architecture

**_Describe the overall architecture of the project. Feel free to add diagrams._**

# Formal Requirements

**_Document that the project fulfills the requirements as documented in the [CNCF graduation criteria for sandbox](https://github.com/cncf/toc/blob/master/process/graduation_criteria.adoc#sandbox-stage)_**

**_Are there any anticipated issues with any of the criteria ?_**

**_Has the TOC been approached for sponsorship?   If yes, which TOC members have agreed to sponsor the project?_**

# CNCF IP Policy

**_Becoming a sandbox project requires adoption of the CNCF IP Policy: [https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy](https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy)_**

**_Note: there is a grace period after becoming a sandbox period to enable projects to adopt the policy, however, some prep is required to ensure there are no major blockers._**

**_Has the IP policy been reviewed?_**

**_List the repos for the project and their current license_**

**_List any dependent repos (upstream/downstream) that are required to build the project  (including but not limited to  libraries, commercial tools, plugins)_**

**_What actions are required to be compliant with the IP policy?_**

# Other Considerations

[...]

PR #355, opened 2020-02-07
Process flowchart, take two

lizrice:
Building on what is in #321, this is an attempt to take the contents of the Google slides version of the project flow, and merge it into our existing process docs. 

I also took the opportunity to eliminate some duplicate (but now out-dated) sandbox criteria documentation


process/project_proposals.adoc
```diff
@@ -1,7 +1,92 @@
...
+All exceptions (and "no" outcomes) are handled by the TOC. Timeframes are approximate, to set expectations.
+
+image::sandbox-process.png[Sandbox process]
+
+. *Project Proposal*
+   * Project proposed through a https://github.com/cncf/toc/issues[GitHub issue]
+   * Template TBD
+. *TOC Triage* _2 weeks max_
+   * Performed by any TOC member
+   * Lazy TOC consensus before declining outright
```
fkautz: any chance this could be worded to convey "not yes" instead of "no"?  I know it sounds trivial, but "not yes" conveys that the project can come back at a later time when it is more mature.

lizrice: good idea!

process/project_proposals.adoc
```diff
@@ -1,7 +1,92 @@
...
+
+=== Sandbox process
+
+All exceptions (and "no" outcomes) are handled by the TOC. Possible "no" outcomes include "not at this time", and the project may be encouraged to re-apply after addressing issues. Timeframes are approximate, to set expectations.
+
+image::sandbox-process.png[Sandbox process]
+
+. *Project Proposal*
+   * Project proposed through a https://github.com/cncf/toc/issues[GitHub issue]
+   * Template TBD
```
xiang90: here is the template: https://github.com/cncf/toc/blob/master/process/project_proposals.adoc

xiang90: oh... this is an editing of the old doc. i guess we should just move https://github.com/cncf/toc/pull/355/files#diff-92faef9f43706befa4fdea2c30f0685bR91-R109 to a new file, and use it as a proposal template.

process/project_proposals.adoc
```diff
@@ -1,7 +1,92 @@
...
+. *TOC Triage* _2 weeks max_
```
xiang90: 2 week max for triage phase is pretty tight. now the pipeline is > 10 weeks. saying "yes" to projects and moving projects to the next phase are straight forward. but if that is not the case, then 2 weeks is not enough since it will involve more discussion between TOCs and the community .

xiang90:
Looks great. thanks @lizrice .

process/project_proposals.adoc
lizrice: There is also https://github.com/cncf/toc/issues/344 in flight. I'd suggest we get this process merged and then we can consolidate on a proposal template as a next step

xiang90: ok. makes sense. 

#344 mainly wants to solve

> Primary aim: This template helps a project to prepare for a presentation to the SIG, and then helps the SIG share the collated information with the TOC.

But I agree with you on consolidation. We should just merge that proposed template with the original project proposal templated since there are a lot of overlappings. Maintainers do not need to create two docs: one for github issue record, one for the sig information collecting.

process/project_proposals.adoc
lizrice: Thanks @xiang90, I have clarified

xiang90:
@lizrice It seems that everyone likes this. probably we can get this merged in the next couple of days if there is no further comments? thanks!

PR #361, opened 2020-02-20
Create Template for Sandbox Projects 

erinaboyd:
This PR addresses  https://github.com/cncf/toc/issues/344”

### DIFF
--- a/sandbox-template.md
+++ b/sandbox-template.md
@@ -1,0 +1,162 @@
+# CNCF SIG Sandbox Project Request Template & Promotion Template
+
+
+# <Project> 
+
+
+## <Promotion/Acceptance> to CNCF <Incubation/Sandbox>
+
+**Primary aim: This template helps a project to prepare for a presentation to the SIG, and then helps the SIG share the collated information with the TOC.**
+
+**Authors:**
+
+
+# Background
+
+**_Link to TOC PR_**
+
+**_Link to Presentation_**
+
+**_Link to GitHub project_**
+
+
+## Project Goal
+
+
+## Current Status
+
+
+## Future Plans
+
+
+# Project Scope
+
+
+## Clear project definition
+
+**_Does the project have a clear and well defined scope_**
+
+## Value-add to the CNCF ecosystem
+
+**_Does the project have a clear value add to the current project ecosystem. How does it relate to other projects with overlapping capabilities._**
+
+
+## Alignment with other CNCF projects
+
+**_Does the project align and actively collaborate with other CNCF projects?_**
+
+**_Does the project require any specific versions of projects (or APIs) to interoperate?  (e.g. K8s API, CSI, CNI, CRI)?_**
+
+**_Does the project augment or benefit other CNCF projects?_**
+
+## Anticipated use cases
+
+**_What problems does the project solve?_**
+
+## Alignment with SIG Reference Model
+
+**_Does the project align with the SIG CNCF reference model and which capabilities does it require/provide at each level of the reference model._**
+
+
+## High level architecture
+
+**_Describe the overall architecture of the project. Feel free to add diagrams._**
+
+# Formal Requirements
+
+**_Document that the project fulfills the requirements as documented in the [CNCF graduation criteria for sandbox](https://github.com/cncf/toc/blob/master/process/graduation_criteria.adoc#sandbox-stage)_**
+
+**_Are there any anticipated issues with any of the criteria ?_**
+
+
+# CNCF IP Policy
+
+**_Becoming a sandbox project requires adoption of the CNCF IP Policy: [https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy](https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy)_**
+
+**_Note: there is a grace period after becoming a sandbox period to enable projects to adopt the policy, however, some prep is required to ensure there are no major blockers._**
+
+**_Has the IP policy been reviewed?_**
+
+**_List the repos for the project and their current license_**
+
+**_List any dependent repos (upstream/downstream) that are required to build the project  (including but not limited to  libraries, commercial tools, plugins)_**
+
+**_What actions are required to be compliant with the IP policy?_**
+
+
+# Other Considerations
+
+## **_Please note, these are not gating criteria but rather to:_**
+
+
+*   **_Collect a standard set of information for each project_**
+*   **_Provides a point in time capture of the state of the project which makes it easier to track progress at future reviews and / or promotion_**
+*   **_Help projects to prepare for SIG and TOC presentation_**
+*   **_Allow the SIG to review the project and perform due diligence for incubation_**
+*   **_Provide the TOC with the information required to accept sponsorship of a project and/or votes_**
+*   **_Identify and rectify any significant issues / blockers prior to presenting to the TOC and acceptance as a CNCF project_**
+
+
+## Cloud Native
+
+**_Does the project meet the definition of Cloud Native?  The CNCF charter states:_**
+
+
+    “Cloud native technologies empower organizations to build and run scalable applications in modern, dynamic environments such as public, private, and hybrid clouds. Containers, service meshes, microservices, immutable infrastructure, and declarative APIs exemplify this approach.
+
+
+    “These techniques enable loosely coupled systems that are resilient, manageable, and observable. Combined with robust automation, they allow engineers to make high-impact changes frequently and predictably with minimal toil.”
+
+
+## Project and Code Quality
+
+**_Are there any metrics around code quality?  Are there good examples of code reviews? Are there enforced coding standards?_**
+
+**_What are the performance goals and results? What performance tradeoffs have been made?  What is the resource cost?_**
+
+**_What is the CI/CD system?  Are there code coverage metrics?  What types of tests exist?_**
+
+**_Is there documentation?_**
+
+**_How is it deployed?_**
+
+**_How is it orchestrated?_**
+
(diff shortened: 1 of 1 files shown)


gerred:
lgtm :+1:

sandbox-template.md
```diff
@@ -0,0 +1,162 @@
...
+## Clear project definition
+
+**_Does the project have a clear and well defined scope_**
+
+## Value-add to the CNCF ecosystem
+
+**_Does the project have a clear value add to the current project ecosystem. How does it relate to other projects with overlapping capabilities._**
+
+
+## Alignment with other CNCF projects
```
k82cn: Should we also include some other projects outside CNCF?
For example, any other similar projects?

sandbox-template.md
```diff
@@ -0,0 +1,162 @@
+# CNCF SIG Sandbox Project Request Template & Promotion Template
```
xiang90: Can we make this merged with project proposal requriements? 

Here is the previous proposal requirements: https://github.com/cncf/toc/blob/master/process/project_proposals.adoc#project-proposal-requirements.

Here is a set of proposals: https://github.com/cncf/toc/tree/master/proposals

So project maintainers only need to create one documentation covering both the SIG needs and the general info as a proposal record.

sandbox-template.md
lizrice: +1 yes please

sandbox-template.md
```diff
@@ -0,0 +1,162 @@
...
+**_What problems does the project solve?_**
+
+## Alignment with SIG Reference Model
+
+**_Does the project align with the SIG CNCF reference model and which capabilities does it require/provide at each level of the reference model._**
+
+
+## High level architecture
+
+**_Describe the overall architecture of the project. Feel free to add diagrams._**
```
ultrasaurus: If we think this should be covered in the project proposal requirements, if we think it is important.  I'd be inclined to just include a presentation outline in process docs and then this is covered by link to video above

sandbox-template.md
```diff
@@ -0,0 +1,162 @@
...
+
+# Formal Requirements
+
+**_Document that the project fulfills the requirements as documented in the [CNCF graduation criteria for sandbox](https://github.com/cncf/toc/blob/master/process/graduation_criteria.adoc#sandbox-stage)_**
+
+**_Are there any anticipated issues with any of the criteria ?_**
+
+
+# CNCF IP Policy
```
ultrasaurus: I would think this section should be in the proposal, where the project would assert the current status and that when additional compliance steps are planned, if accepted

sandbox-template.md
```diff
@@ -0,0 +1,162 @@
...
+**_List any dependent repos (upstream/downstream) that are required to build the project  (including but not limited to  libraries, commercial tools, plugins)_**
+
+**_What actions are required to be compliant with the IP policy?_**
+
+
+# Other Considerations
+
+## **_Please note, these are not gating criteria but rather to:_**
+
+
```
ultrasaurus: This section seems like good guidance to a SIG chair (or their delegate) who is reviewing the project and preparing information for TOC recommendation, but doesn't seem link it belongs in template.  (Maybe move to top and include in a section which is "guidance: to be removed when using this template"

sandbox-template.md
```diff
@@ -0,0 +1,162 @@
...
+**_Does the project meet the definition of Cloud Native?  The CNCF charter states:_**
+
+
+    “Cloud native technologies empower organizations to build and run scalable applications in modern, dynamic environments such as public, private, and hybrid clouds. Containers, service meshes, microservices, immutable infrastructure, and declarative APIs exemplify this approach.
+
+
+    “These techniques enable loosely coupled systems that are resilient, manageable, and observable. Combined with robust automation, they allow engineers to make high-impact changes frequently and predictably with minimal toil.”
+
+
+## Project and Code Quality
```
ultrasaurus: Parts of this are covered in CII best practices.  Would be great to ask projects to fill out specific questions so this can be collected in standard format (that will then be linked from landscape)

sandbox-template.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Open Governance
+
+**_How are committers chosen?_**
+
+**_How are architectural and roadmap decisions made?_**
+
+**_How many decision makers are outside the sponsoring organization._**
+
+## Adoption
```
ultrasaurus: This section seems like it should be in the proposal -- the project would be answering this, rather than SIG

sandbox-template.md
```diff
@@ -0,0 +1,162 @@
...
+
+**_Does the project participate in a CNCF User Group?_**
+
+## Vendor Independence
+
+**_Is the project reasonably independent from the sponsoring vendor?_**
+
+**_Are all communication channels and project resources hosted just for this project or with other CNCF projects/resources?_**
+
+**_Is all code that is part of the project hosted and part of the CNCF managed orgs and repos?_**
```
ultrasaurus: Wouldn't the answer always be "no" for this?  since the project is not yet accepted into the CNCF?

lizrice: Suggest rephrasing to check that the intention is to transfer all of these to the CNCF

lizrice:
Let me start by apologising, I was very confident about this template earlier in the week on the TOC SIG-chairs call but taking a closer look at it, I'm a bit confused about exactly what we're talking about here. I'm sorry that I may well have created confusion for everyone else as well, especially as I've maybe not always been very rigorous in using "PR" when I mean "issue" and vice versa. 

In the [process](https://github.com/cncf/toc/blob/master/process/project_proposals.adoc#sandbox-process) we start with an *issue* raise by the project. 

Implied by history and by the process, but not actually spelled out, there needs to be a project proposal document in a *PR* that would add a document to [`proposals/sandbox`](https://github.com/cncf/toc/tree/master/proposals/sandbox) when merged. This document includes the items listed [here](https://github.com/cncf/toc/blob/master/process/project_proposals.adoc#project-proposal-requirements) and is the historical record of the project proposal. 

The issue should link to that PR, but it doesn't need to duplicate what's listed in the proposal. 

@erinaboyd is this template intended to be used for the issue, or for the PR?

erinaboyd:
@lizrice so, great question. I don't think projects need both issue and PR. I think it's an opportunity to streamline our process. The intention of the template is to gather data to start an evaluation that can first be done by the SIGs then moved to the TOC after the project has been reviewed. This should then be used to build on going to the next level (sandbox->incubation, incubation->graduation).
Historically, I thought we started with issues, but then found PRs to be a better fit to evaluate projects. Issues (imo) were just to create a request with very little information in them and used as a means to get the project on the TOC schedule (when it used to be a single process)

resouer:
If I understand correctly, this PR is the template for SIG recommendation workflow. 

@erinaboyd Could you please refactor the PR according to this example: https://github.com/cncf/sig-network/pull/3/files, per recommended by @caniszczyk here: https://github.com/cncf/toc/issues/380#issuecomment-594733367

I'd suggest change the PR title to "Template for SIG recommendation for Sandbox projects" if it's the case.

raravena80:
+1, I think it would help to clearly differentiate between "Template for SIG recommendation for Sandbox projects" (Usually created by SIGs) and "Template for Sandbox project donation" (Usually created by the project members/maintainers).

xiang90:
@lizrice @erinaboyd 

As I mentioned here: https://github.com/cncf/toc/pull/361#discussion_r382741634, the project donation proposal requirements and this SIG recommendation requirement overlap quite a bit. 

We can either just combine them into one or make the donation one lighter. There is no need for the duplication.

erinaboyd:
@lizrice @raravena80 @xiang90 So, I thought the whole purpose of this was to create 1 template to gather information for SIG review and this be passed onto the TOC. It would include everything you needed for donation and it would be consistent between SIGs.

The ideas is to consolidate and create a 'standard'. Please correct me if I am wrong here.

raravena80:
@erinaboyd I originally thought that it was 1 template too. But IIRC on one of the TOC-Chairs meetings, there was a discussion about having a template for the project to submit/donate, authored by the project submitters and then another 'recommendation/or not' template authored by the SIG chairs/tech leads.

For Sandbox, I believe the natural thing is for the SIG to ask a given project to provide all the information initially and then have the SIG drive the recommendation after verification.

The question in my view would be whether that is either one of these (or other alternatives):

#### Single template
1. A single recommendation github PR+doc (authored by the SIG).
   1. The PR file could go under `recommendations/` in the SIG git repo.
   1. The SIG will gather project information offline using their individual methods.
2. Combination of gdoc from the project (authored by project submitters), and recommendation github PR+doc (authored by the SIG).
   1. The PR file could go under `recommendations/` in the SIG git repo.
3. The SIG lets the project submitters create the PR with a single file under `recommendations/` in the SIG git repo and the SIG reviews/requests changes/merges the PR. In this case, the authors would be someone from the project and someone from the SIG.

#### Two templates

1. Combination of github PR+doc from the project (authored by project submitters), and recommendation github PR+doc (authored by the SIG).
   1. One file could go under `recommendations/` and another file could go under `proposals/` in the git repo.

jberkus:
Speaking as someone who shepherds some projects through this process, I would vastly prefer a workflow that only involves a single template.

xiang90:
> Speaking as someone who shepherds some projects through this process, I would vastly prefer a workflow that only involves a single template.

I agree. I think we should consolidate into one template. That template contains enough information for both the donation needs and the SIG reviews/recommendation.

amye:
Closing as this is replaced by the form submission process. 
Many thanks to all who have been working to improve our sandbox processes!

PR #381, opened 2020-03-04
Update README.md

raravena80:
Updating SIG-Runtime and date.


raravena80:
@amye @caniszczyk seems like a straight forward change. Let me know if you need anything else. Thx.

PR #388, opened 2020-03-13
Clarify dependencies

lizrice:
Following TOC discussion I'm suggesting some clarification around the acceptability of dependencies

### DIFF
--- a/PRINCIPLES.md
+++ b/PRINCIPLES.md
@@ -78,6 +78,8 @@ Via the “no kingmakers” principle and “what is a project”, the CNCF may 
 
 This means that the CNCF is not promoting a single, monolithic stack of technologies. 
 
+CNCF projects may have dependencies on other open source projects, which may be outside the CNCF. Widely-used dependencies are preferred; pluggable dependencies fit well with our principles but are not essential.  _Examples: Kubernetes' dependency on etcd, which was not a CNCF project at the time Kubernetes was brought into the CNCF; CNI, which enables pluggable networking dependencies._ 
+
 The CNCF is a badge of quality and velocity. CNCF projects should be on a path to being tools that users can trust, broadly work together, and that meet other cloud native criteria. But the CNCF badge does not mean “this is the standard tool”.
 
 - Overlapping projects are ok, especially where they make significantly different design tradeoffs
--- a/PRINCIPLES.md
+++ b/PRINCIPLES.md
@@ -78,7 +78,7 @@ Via the “no kingmakers” principle and “what is a project”, the CNCF may 
 
 This means that the CNCF is not promoting a single, monolithic stack of technologies. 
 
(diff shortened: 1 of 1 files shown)


PRINCIPLES.md
```diff
@@ -78,6 +78,8 @@ Via the “no kingmakers” principle and “what is a project”, the CNCF may
 
 This means that the CNCF is not promoting a single, monolithic stack of technologies. 
 
+CNCF projects may have dependencies on other open source projects, which may be outside the CNCF. Widely-used dependencies are preferred; pluggable dependencies fit well with our principles but are not essential.  _Examples: Kubernetes' dependency on etcd, which was not a CNCF project at the time Kubernetes was brought into the CNCF; CNI, which enables pluggable networking dependencies._
```
caniszczyk: You may want to mention that "CNCF projects may have dependencies on other open source projects that comply with the CNCF IP policy: https://github.com/cncf/foundation/blob/master/whitelist-policy.md"

justincormack: This clarification was referring to external dependencies not in the project codebase, such as Postgres for Harbor, so the IP policy does not apply. Maybe the wording needs to make this clearer.

lizrice: I think the IP policy does apply to dependencies (from my reading of it, but open to correction!)

justincormack: I read that as the IP policy for pieces that are contained in the project, https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy refers to inbound and outbound code and not to runtime dependencies (eg use of Amazon S3 was brought up in the discussion as being a common use case).

VinodAnandan:
I think it's better to verify the OSS license compliance with all existing CNCF projects and see the practicality. I also think it's better to grant some time for existing and new projects to achieve the license compliance.

Few open questions

- How deep the dependency license check should be carried out?
- What if there are transitive dependencies without any license (unlicensed code)?

lizrice:
@VinodAnandan this is something the CNCF legal team already do for projects coming into the CNCF

PRINCIPLES.md
lizrice: I think there's a difference here for required and optional dependencies. An option to use Amazon S3 might be very useful for lots of the community, but other options should be supported. I don't think CNCF projects can have a _required_ dependency (run or build time) on just one particular vendor platform. 

As another example, the Postgres license is supported in the IP policy so that projects can rely on it, but it's up to the user to decide whether to use the OSS or a managed version.

lizrice: @caniszczyk please could you check whether what I suggested here is accurate?

caniszczyk: having a dependency on one particular vendor would get against our values 3(g) and is up to the discretion of CNCF to call this out if we find it: https://github.com/cncf/foundation/blob/master/charter.md#3-values

for example, rook in the beginning only supported say ceph but expanded to support other storage systems (they always had the plug-in infrastructure in place).

S3 is an interesting case as there is the S3 API vs AWS S3 service delineation

PR #402, opened 2020-03-25
Update SIG liaisons in sig specific md files

alena1108:
@lizrice noticed that sigs have liaisons mentioned in their respective md files; updated these files based on https://github.com/cncf/toc/pull/362


xiang90:
lgtm

PR #403, opened 2020-03-25
Remove obsolete hard-coded TOC members count

alena1108:
Add a link to the corresponding github page as a reference instead. Github link if more members specific than the original https://www.cncf.io/people/technical-oversight-committee/ link, but I'd be happy to put the original reference back if the PR reviewer decides otherwise.


caniszczyk:
This is reasonable thanks @alena1108

Issue #401, opened 2020-03-23
Proposal: Artifact Hub as Sandbox project


mattfarina:
The [current process](https://github.com/cncf/toc/blob/de7446cd7d9fcb2b9516b675a5b061c9eeea69ac/process/project_proposals.adoc) notes creating an issue to propose a project for sandbox. This issue serves as that step.

A pull request with more details will be coming soon that includes the usual information and more detail on the proposal.

This is for https://artifacthub.io/ whose code is at https://github.com/artifacthub

michelleN:
Did we get a decision on whether this would go under SIG-Runtime or SIG App Delivery? @lizrice

quinton-hoole:
I think realistically both of those sigs will want to take a close look.  Also, is that GitHub repo still empty?

mattfarina:
@michelleN I'm curious why SIG Runtime. SIG App Delivery I can see because of operators and Helm. SIG Security I could see because of Falco and OPA (which have been discussed). I'm curious of the reason for SIG Runtime.

quinton-hoole:
There seems to be a fair amount of overlap with container registries, which fall within SIG Runtime.    See also https://github.com/cncf/toc/blob/master/sigs/runtime-charter.md#interfaces-with-other-related-groups

raravena80:
I believe the repo is https://github.com/artifacthub/hub. There also seems to be a DB component that might fall within SIG-Storage: https://github.com/artifacthub/hub/tree/master/database.

I think from the SIG-Runtime pov: Helm Generates YAML template -> Template contains pods -> Pod definition contains container definitions -> container points to container registry -> Images are pulled from container registry -> Containers created/started. My guess is that this would not apply to OPA policies and Falco rules.

Another aspect would also be checking the validity of the container images that Helm charts refer to.

xiang90:
@michelleN 

I think the main SIG responsible for this should be SIG App Delivery. SIG App Delivery should decide which other SIGs need to be involved and coordinate the review process accordingly.

I am going to assign this to SIG App Delivery if I hear no objection.

xiang90:
@mattfarina I assigned to SIG App Delivery for the initial review. Note that the Sandbox proposal still needs to be added. The template is TBD as it is now (the old one looks like [this](https://github.com/cncf/toc/blob/de7446cd7d9fcb2b9516b675a5b061c9eeea69ac/process/project_proposals.adoc#project-proposal-requirements)). I assume it should be ready in the next couple of weeks.

[...]

PR #409, opened 2020-04-01
Adding Artifact Hub sandbox proposal

mattfarina:
The Artifact Hub is a hub for cloud native projects to have their
artifacts listed and easily found by end-users. The goal of the
Artifact Hub is easy discovery of cloud native artifacts
(e.g., Helm charts, Falco rules, etc) and to provide information
about the artifacts so that end-users can make decisions about
using them.

Related issue #401

### DIFF
--- a/proposals/sandbox/artifacthub.md
+++ b/proposals/sandbox/artifacthub.md
@@ -1,0 +1,160 @@
+# Artifact Hub Sandbox Proposal
+
+## Project name and description
+
+The [Artifact Hub](https://artifacthub.io/) is a hub for cloud native projects to have their artifacts listed and easily found by end-users. The goal of the Artifact Hub is easy discovery of cloud native artifacts (e.g., Helm charts, Falco rules, etc) and to provide information about the artifacts so that end-users can make decisions about using them.
+
+short name: artifacthub
+
+## Alignment with CNCF charter mission
+
+The [CNCF charter](https://github.com/cncf/foundation/blob/84dc259f50b89d656907f2247d9415d8c10c3efd/charter.md) notes that:
+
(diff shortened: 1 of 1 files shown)


mattfarina:
Questions and Suggests from the SIG App Deliver meeting on 4/1. Note, this is a sandbox proposal: 

* We were asked to have more clear contributor docs and details around the project.
* It was suggested by Alois that we start doing the incubator due diligence and show how it's on a path to meet that. Note, this is for a sandbox submission
* Deeper presentation on Artifact Hub requested. How it relates to other CNCF projects, roadmap, etc. This will be scheduled when more of the questions asked can be accounted for.
* Having contributors from multiple orgs was requested. It's not clear if this contributors or committers from multiple orgs. Committers from multiple orgs is a graduation requirement.
* The artifact hub is listed as pre-alpha. They are looking for tagged releases as part of the project which I would assume comes in at alpha or beta. We may need to seek clarity on what early stage means according to [sandbox project documentation](https://github.com/cncf/toc/blob/master/process/sandbox.md) which do call out experimental projects.
* Some elements in the proposal were asked for more clarity. Such as the goal. It currently sits at a sentence but it was asked to be expanded.
* The question, "why should this be part of the CNCF?" was asked to be answered. This is not one of the questions currently being asked as part of the templates and would be great to be added.

I imagine I missed some other elements. If there are things that need to be addressed somewhere please list them here and we can attempt to address them.

caniszczyk:
ArtifactHub was accepted into the sandbox:
https://lists.cncf.io/g/cncf-toc/message/4853

project onboarding will happen here https://github.com/cncf/toc/issues/470

PR #432, opened 2020-04-21
Align individual sigs pages info with README

alena1108:
Chairs section in particular

cc @amye

### DIFF
--- a/sigs/storage-charter.md
+++ b/sigs/storage-charter.md
@@ -116,10 +116,10 @@ This SIG follows the [standard operating
 guidelines](https://github.com/cncf/toc/blob/master/sigs/cncf-sigs.md#operating-model)
 provided by the TOC unless otherwise stated here.
 
-**Current TOC Liaison:** Xiang Li
+**TOC Liaison:** [Xiang Li](https://github.com/xiang90)
 
-**[Proposed] Co-Chairs:** Alex Chircop, Quinton Hoole, “One Other TBD”
+**SIG Chairs:** [Alex Chircop](https://github.com/chira001), [Quinton Hoole](https://github.com/quinton-hoole), [Erin Boyd](https://github.com/erinboyd)
 
-**[Proposed] Tech Leads:** Saad Ali, Xing Yang, “One Other TBD”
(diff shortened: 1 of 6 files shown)


sigs/storage-charter.md
```diff
@@ -116,10 +116,10 @@ This SIG follows the [standard operating
...
 provided by the TOC unless otherwise stated here.
 
-**Current TOC Liaison:** Xiang Li
+**TOC Liaison:** [Xiang Li](https://github.com/xiang90)
 
-**[Proposed] Co-Chairs:** Alex Chircop, Quinton Hoole, “One Other TBD”
+**SIG Chairs:** [Alex Chircop](https://github.com/chira001), [Quinton Hoole](https://github.com/quinton-hoole), [Erin Boyd](https://github.com/erinboyd)
 
-**[Proposed] Tech Leads:** Saad Ali, Xing Yang, “One Other TBD”
+**Tech Leads:** Saad Ali, Xing Yang, “One Other TBD”
```
alena1108, line 123: @amye this section might need adjustments too - Saad Ali being a TOC now. @chira001 do we have an updated list, or should I put "Xing Yang, 2 TBD"?

chira001: Hmmm - this is outdated - the actual list is at current list is at: https://github.com/cncf/sig-storage

Also - we have a copy of the charter in the SIG repo: https://github.com/cncf/sig-storage/blob/master/storage-charter.md  (as do the other SIGs - I believe)

This is confusing - should the SIG charters be mastered in the individual SIG repos?  (that would be my preference)

raravena80: +1 for charters to be in a single place. Sounds like an agenda item for the next TOC-Chairs meeting 😄

alena1108: @chira001 @raravena80 for now, I'll go ahead and do the minimal updates here (the chairs)

Agree that we need to have a bigger discussion on the charters placement to avoid duplication/inconsistencies. I can see benefits of having some minimal information on SIGs and their mission statement in cncf/toc repo too, as an initial reference.

PR #448, opened 2020-05-14
Shift to a one week comment period

caniszczyk:
Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>


lizrice:
I think there were sufficient concerns about this on the mailing list that we should leave it as is

caniszczyk:
@lizrice agreed, I'll try again in 6-12 months, I think we have gone overboard on the comment period given that the work is all done in the open to begin with

PR #446, opened 2020-05-08
SIG member roles: TOC Liaison role description and responsibilities

alena1108:
cc @lizrice

PR #450, opened 2020-05-18
Follow-up fix on TOC liaison definition

alena1108:
To address @lizrice suggestions that didn't make it to the original PR https://github.com/cncf/toc/pull/446


quinton-hoole:
LGTM

Issue #478, opened 2020-06-24
[PROJECT ONBOARDING] Cloud Custodian


amye:
Welcome to CNCF Project Onboarding!

Please ensure that you:
- [x] Understand the project proposal process and reqs: https://github.com/cncf/toc/blob/master/process/project_proposals.adoc#introduction
- [x] Understand the services available for your project at CNCF https://www.cncf.io/services-for-projects/
- [x] Ensure your project meets the CNCF IP Policy: https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy
- [x] Has your project adopted open governance already? see http://opengovernance.dev

Formal onboarding:
- [x] Domain: transfer domain to CNCF/LF (ITx)
- [ ] Trademarks: transfer any trademark and logo mark assets over to the LF
- [x] GitHub: ensure 'thelinuxfoundation' and 'caniszczyk' are added as initial org owners
- [x] Artwork: Ensure logos present on https://github.com/cncf/artwork
- [x] Website: ensure LF footer is there and website guidelines followed, analytics transferred
- [x] Create maintainer list + added to aggregated https://maintainers.cncf.io list - OWNERS file needed
- [x] Devstats: add to devstats https://devstats.cncf.io/
- [x] Marketing: update relevant intro + slide decks
- [x] Trail Map: update CNCF trail map if relevant (incubating projects and above)
- [x] Update https://landscape.cncf.io
- [x] Events: update CFP + Registration + CFP Area forms
- [x] ServiceDesk: confirm maintainers have read https://www.cncf.io/services-for-projects/
- [x] CNCF Welcome Email Sent to confirm maintainer list access
- [x] Schedule monthly project sync with CNCF staff

amye:
Adding @johnmark @kapilt from the project side.

amye:
@johnmark, @kapilt: is there an OWNERS file for you? Once I have that, I can add you to maintainers.cncf.io.

amye:
Added to https://www.cncf.io/sandbox-projects/ with the logo that was available on the Cloud Custodian site, let me know if you have more links to add.

amye:
@johnmark, @kapilt: checking in on the OWNERS file, thanks!

amye:
Closing, as trademarks are in progress offline.

PR #480, opened 2020-06-25
Proposing Cloud Custodian to CNCF as sandbox project

johnmark:


amye:
Cloud Custodian was accepted into the sandbox:
https://lists.cncf.io/g/cncf-toc/message/4853

project onboarding will happen here https://github.com/cncf/toc/issues/478

PR #499, opened 2020-07-28
Updating with new sandbox form

amye:

PR #505, opened 2020-08-05
Updates for new sandbox application process

lizrice:
Needs links to the form and spreadsheet to be added


lizrice:
@amye could you use the review suggestion feature to add the links to the Sandbox application form and public spreadsheet please?

lizrice:
Oh wait, I see you already have something under #499. Ignore this. I'll add some comments into that instead

PR #509, opened 2020-08-11
Update due-diligence-guidelines.md to be explicit about dependencies

chira001:
Be explicit about collecting dependencies and listing them

### DIFF
--- a/process/due-diligence-guidelines.md
+++ b/process/due-diligence-guidelines.md
@@ -82,7 +82,7 @@ the detail where necessary.
 * What are the most important holes? No High-Availability? No flow control? Inadequate integration points?
 * Code quality.  Does it look good, bad or mediocre to you (based on a spot review).  How thorough are the code reviews? Substance over form.
   Are there explicit coding guidelines for the project?
-* Dependencies.  What external dependencies exist, do they seem justified?
+* Dependencies.  What external dependencies exist, do they seem justified?  Note: all core dependencies should be listed in the document along with the details of relevant repos
 * What is the release model?  Versioning scheme?  Evidence of stability or otherwise of past stable released versions?
 * What is the CI/CD status?  Do explicit code coverage metrics exist? If not, what is the subjective adequacy of automated testing?
   Do different levels of tests exist (e.g. unit, integration, interface, end-to-end), or is there only partial coverage in this regard?  Why?


chira001:
@lizrice @quinton-hoole @erinaboyd - making a short addition to the DD guidelines to make documenting dependencies an explicit requirement

process/due-diligence-guidelines.md
```diff
@@ -82,7 +82,7 @@ the detail where necessary.
 * What are the most important holes? No High-Availability? No flow control? Inadequate integration points?
 * Code quality.  Does it look good, bad or mediocre to you (based on a spot review).  How thorough are the code reviews? Substance over form.
   Are there explicit coding guidelines for the project?
-* Dependencies.  What external dependencies exist, do they seem justified?
+* Dependencies.  What external dependencies exist, do they seem justified?  Note: all core dependencies should be listed in the document along with the details of relevant repos
```
dims, line 85: "core" implies "non-optional"?

chira001: yes - that was the intent - happy to change it if there is better wording.

dims: "non-optional" seems better otherwise folks will debate what "core" is/means :)

dims:
@chira001 do we want to add another list item about cross checking licenses for all dependencies and make a list of things that do are not white listed here? https://github.com/cncf/foundation/blob/master/allowed-third-party-license-policy.md

( this is a slightly different problem then the original above )

process/due-diligence-guidelines.md
lizrice: I don't know how to get the terminology right here but I think we need to distinguish between significant functional dependences, and programming language / package dependencies. We don't want to be reviewing giant lists of, say, Go packages or Node dependencies, but we do want to know when a project assumes the existence of a significant external component.

chira001:
> @chira001 do we want to add another list item about cross checking licenses for all dependencies and make a list of things that do are not white listed here? https://github.com/cncf/foundation/blob/master/allowed-third-party-license-policy.md
> 
> ( this is a slightly different problem then the original above )

I think this is already covered by "Does it have a license? Which one? Does it have a CLA or DCO? Are the licenses of it's dependencies compatible with their usage and CNCF policies? CNCF staff will handle the full legal due diligence."  and the graduation policy already specifies that the project needs to comply with the CNCF IP Policy, but you are right - it would be helpful to include a reference in the guidelines.

erinaboyd:
@chira001 @lizrice really what we are trying to get at though is "does it have a non-open source 3rd party license?"
Maybe that would be easier listing than all the possible licenses

dims:
@erinaboyd we already have a list :) https://github.com/cncf/foundation/blob/master/allowed-third-party-license-policy.md

PR #529, opened 2020-09-03
Update CONTRIBUTING.md

rtatnet3:
Updating for PR > issues


amye:
We're actually moving away from having projects put things into this repo as part of the proposal process, closing without merge.

PR #551, opened 2020-10-16
Added requirement to have basic contributor information.

jberkus:
This is listed as required on the sandbox intake form, but it was not listed
as required on the process doc for sandbox projects.

As such, I've added it to the process doc.  If things should have gone the other way (i.e. it's a mistake for it to be included on the form), then please fix the form and reject this.

cc @amye   @lizrice

### DIFF
--- a/process/graduation_criteria.adoc
+++ b/process/graduation_criteria.adoc
@@ -13,6 +13,7 @@ To be accepted in the sandbox a project must
 * Require 3 of the 11 TOC members to step forward as sponsors to enter the sandbox
 * Adopt the CNCF https://github.com/cncf/foundation/blob/master/code-of-conduct.md[Code of Conduct]
 * Adhere to CNCF https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy[IP Policy] (including trademark transferred)
+* Have a basic CONTRIBUTING.md file or other information on how to contribute to the project 
 * List their sandbox status prominently on website/readme
 
 See the https://github.com/cncf/toc/blob/master/process/sandbox.md[CNCF Sandbox Guidelines v1.0] for the detailed process.


process/graduation_criteria.adoc
```diff
@@ -13,6 +13,7 @@ To be accepted in the sandbox a project must
 * Require 3 of the 11 TOC members to step forward as sponsors to enter the sandbox
```
lizrice, line 13: This line should be deleted, it's no longer true

process/graduation_criteria.adoc
```diff
@@ -13,6 +13,7 @@ To be accepted in the sandbox a project must
...
 * Adopt the CNCF https://github.com/cncf/foundation/blob/master/code-of-conduct.md[Code of Conduct]
 * Adhere to CNCF https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy[IP Policy] (including trademark transferred)
+* Have a basic CONTRIBUTING.md file or other information on how to contribute to the project
```
lizrice, line 16: I don't honestly think we need this - there's no need to duplicate everything that's in the form, it will just get out of sync (and for example we are also asking for sandbox projects to have some form of public roadmap, but that's not listed here). 

You could argue that in that case, nothing should be listed here, but I think having the big ticket items here is helpful. We want to make sure it's absolutely clear that, for example, the project is going to have to transfer IP to CNCF.

jberkus: I disagree; forms are a very poor way for projects to find out what's expected of them.  What happens is that the project takes care of the items listed on Requirements, gets to the form, and then is completely baffled.  The form is not documentation, and springing surprises on submitting projects isn't nice and re-enforces the "insider" nature of the process.

We're also working on a checklist for projects (https://github.com/cncf/sig-contributor-strategy/pull/59); it's possible that maybe Requirements should be only the big items and the checklist should have all the detailed items.  That would mean having the requirements doc link out to the checklist, once it's finalized.

jberkus: In fact ... I wasn't aware of the roadmap requirement until you mentioned it here, and I'm hardly a stranger to the process.

dzolotusky: I agree that the form isn't documentation, but I'm even more concerned about having many different sources of truth for this information all under the CNCF org in Github. I feel like that would make the "insider" problem even worse. Rather than creating a new file under sig-contributor-strategy, I think that we should do our best to have 1 source for this information and link to it from all of the places that need it.

If there's no way to keep the form and a .md file in sync automatically, maybe we just invest in giving every field in the form a good description and turn the form into the documentation and single source of truth for this.

jberkus: It's not hard to keep an MD file and a form in sync, especially for a form that short.  The other problem with form-as-doc is that the form is not maintainable via a PR/discussion process, they way the requirements are (and we're demonstrating there).  And agreed that it would be better to have all of the requirements in one file.

dzolotusky: If you have a way to keep an MD file and a form in sync, could you propose whatever that mechanism is here? I feel like that's the clear best answer for this.

jberkus: this is now a much larger TODO.  I'll keep it open until that's expressed, then close the PR

jberkus:
Closing in favor of issue #554

PR #568, opened 2020-11-13
PROJECTS.csv added

idvoretskyi:
PTAL @caniszczyk @amye 

Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>


caniszczyk:
Thanks @amye will update this over time so it's all in one place

PR #579, opened 2020-12-15
Add "can" to cloud native definition

xmulligan:
https://cloud-native.slack.com/archives/C01F1LVAQCC/p1607974863215600
see thread for reference in the telco space. The main premise is that just because something is using containers or microservices does not automatically make it cloud native. This update would help clarify that those techniques can help make things cloud native, but do not automatically do so.


electrocucaracha:
Usually examples help to clarify definitions but in this case I'm not sure if adding `can` will help on that. I'm agree that the current definition is accurate but it can be interpreted differently.

PR #592, opened 2021-01-26
Update project-onboarding.md

amye:
Editing for clarity


caniszczyk:
@amye can you add something about if they have an event, say like 'KubeCon' or 'Crossplane Day' that the event moves over to CNCF for management, that's something we should be upfront about so we don't miss it

amye:
This is really meant for sandbox projects, so let me think about how that might be meaningful.

caniszczyk:
@amye I'll merge but we still should have a bullet about transferring events over as it's come up multiple times

PR #601, opened 2021-02-01
Update French version

joellord:
États de l'art was a literal translation of State-of-the-art and is not generally accepted in French. "La fine pointe" would be a better match.


caniszczyk:
@joellord thanks!

PR #610, opened 2021-02-24
Add requirement for documented security processes

lizrice:
As discussed in TOC meeting Feb 16. The idea is not to impose any particular security processes, just to clarify that they must be in place. 
I'm proposing that this should be in place at Incubation level, since we expect it to be in production with some users at that stage.

See also 
* https://github.com/cncf/sig-security/issues/538 where SIG Security have offered to assess whether existing incubated / graduated projects have process in place
* https://github.com/cncf/sig-security/issues/183 on providing a template - TOC doesn't want to *impose* a specific process, but guidance for projects would be helpful

### DIFF
--- a/process/graduation_criteria.adoc
+++ b/process/graduation_criteria.adoc
@@ -29,6 +29,7 @@ To be accepted to incubating stage, a project must meet the sandbox stage requir
  * Demonstrate a substantial ongoing flow of commits and merged contributions.
  * Since these metrics can vary significantly depending on the type, scope and size of a project, the TOC has final judgement over the level of activity that is adequate to meet these criteria
  * A clear versioning scheme.
+ * Clearly documented security processes explaining how to report security issues to the project, and describing how the project provides updated releases or patches to resolve security vulnerabilities 
  * Specifications must have at least one public reference implementation.
 
 === Graduation Stage


process/graduation_criteria.adoc
```diff
@@ -29,6 +29,7 @@ To be accepted to incubating stage, a project must meet the sandbox stage requir
  * Demonstrate a substantial ongoing flow of commits and merged contributions.
  * Since these metrics can vary significantly depending on the type, scope and size of a project, the TOC has final judgement over the level of activity that is adequate to meet these criteria
  * A clear versioning scheme.
+ * Clearly documented security processes explaining how to report security issues to the project, and describing how the project provides updated releases or patches to resolve security vulnerabilities
```
dims, line 32: s/processes/process/

lizrice: I think these could be more than one process - one for 3rd party to tell them how to report new a security issue, another for documenting how project responds when vulnerabilities are found

dims: sounds good

dims, line 32: Could we please showcase some existing / widely used process for a minimum bar here? For example the one from GH itself:

https://docs.github.com/en/code-security/security-advisories/adding-a-security-policy-to-your-repository

dims: when the sig-security folks write up their recommendations, we can switch the url around to what they come up with.

dims: i am fine with this merging as-is

caniszczyk:
let's leave this open until Friday morning, ping @cncf/toc for their +1s

dims:
+1 from m.e

lizrice:
not sure if I need to +1 my own PR but here it is 😀

saad-ali (commented):
/lgtm

+1

rochaporto:
+1

dzolotusky:
+1

justincormack:
+1

alena1108:
+1

erinaboyd:
+1

PR #653, opened 2021-05-04
Added Hebrew Translation

ErezYalon:


caniszczyk:
LGTM after putting it through google translate, thank you!

PR #656, opened 2021-05-09
Added Turkish option to definition

canberkkoc:
Added Turkish option


caniszczyk:
thank you @canberkkoc, looks OK via google translate :)

PR #664, opened 2021-05-25
Proposing Krustlet for Sandbox

thomastaylor312:
This is the proposal for adding Krustlet as a sandbox project. As per the documentation, the Google Form has also been filled out. Please let me know if there is any missing information in the proposal and I can go back and add it!

PR #665, opened 2021-05-25
fix(process): Removes PR step for sandbox projects

thomastaylor312:
I found out in #664 that you only need to submit the form and not a PR.
However, it does seem like the sandbox diagram and the documentation have
drifted out of sync with each other and I wasn't sure enough of the
process to update that


caniszczyk:
@amye we should leave this but the project can PR for record keeping purposes AFTER being accepted, I think it's useful to have all the info in one place

amye:
Want to put this as an 'after acceptance' step? That solves the problem of not having in Github.

PR #666, opened 2021-05-25
Add Portuguese-PT translation.

JoaoDuarteGomes:
Update pt-BR link description to prevent confusion with pt-PT.


pasimoes (commented):
This text is exactly the same as PT-BR. Do we need really have both?

ErezYalon:
It is not exactly the same, but similar. Why not have both?

PR #698, opened 2021-07-19
Update projects.md

amye:
Correcting an error: kubevirt is sandbox, kubeedge is incubating.


caniszczyk:
@amye make sure PROJECTS.csv is updated

amye:
@caniszczyk: .csv was right, .md was the one in error

PR #785, opened 2022-02-14
Update README.md to correct storage co-chairs

chira001:
Correct Storage co-chairs


chira001:
@amye - for review - thanks!

PR #834, opened 2022-05-06
Add Thai translation to DEFINITION

zdk:
Thai definition


caniszczyk:
thanks, passes the google translate sniff test :)

PR #847, opened 2022-06-08
Update links from .adoc to .md

carolynvs:
Previously we had renamed some pages from .adoc to .md, which broke some links. I've updated the links for the proposal_process.adoc and graduation_criteria.adoc pages to use the new extension, .md. I have also done a sweep of the entire repository for those old links and updated previously submitted reviews and proposals so that their links work too.


amye:
Much appreciated!

PR #849, opened 2022-06-10
docs: projects: fix link to CRI-O announcement

DavidKorczynski:
The old link was set to the etcd announcement. We should use the CRI-O announcement.

@caniszczyk 

Signed-off-by: DavidKorczynski <david@adalogics.com>


caniszczyk:
@DavidKorczynski thanks I'm an idiot ;)

PR #913, opened 2022-08-26
Update serverless WG info

duglin:
Signed-off-by: Doug Davis <dug@microsoft.com>


dims:
This PR reflects status quo. This WG is active! so yay!!

dims:
thanks a ton @duglin !

PR #932, opened 2022-09-27
Add TAG Environmental Sustainability to README and order list

leonardpahlke:
Signed-off-by: leonardpahlke <leonard.pahlke@googlemail.com>

This PR updates the README
* Add link to environmental sustainability TAG
* List of TAGs rearranged in alphabetical order.


leonardpahlke:
cc @amye

amye:
The order was actually 'in the order they were created', so let me think if now's the right time to change that.

PR #960, opened 2022-11-15
Improve Portuguese-BR definition

rmleme:
- Minor grammatical improvements and fixes (_e.g._ "ecosistema" --> "ecossistema"; "Containeres" --> "Contêineres"; "cloud native" is already consolidated in Portuguese-BR, no need for "Tecnologias nativas ao cloud").
- Portuguese-BR was made closer to Portuguese-PT (the difference between the two variants of Portuguese language should be minimal, just like en-US and en-GB).


krol3:
@resouer could you help us with the review?

PR #1012, opened 2023-02-17
armadaproject org change

kannon92:
G-Research/armada is now armadaproject/armada.  Change the docs to reflect this!


nate-double-u (approved):
Thanks for this update @kannon92

Issue #992, opened 2023-01-11
Add clarification to CNCF graduation criteria


cathyhongzhang:
Current CNCF graduation criteria do not lay out everything the TOC looks at. For example, TOC will ask the project leads to reach out to TAG-security for a security check and get their recommendation/input, but this is not documented in the graduation criteria or graduation DD. TOC is also discussing whether to leverage TAG Contributor Strategy's expertise to help review the project's governance to make sure it is open and neutral. Without clearly spelling out these, there could be confusion among the project maintainers. We have seen some stumbling along this application journey. To set the project maintainers up well for the graduation review process and avoid potential insistency/controversy about the evaluation/approval/rejection result, it is worth adding such clarification to the graduation criteria/DD document. Thoughts?

TheFoxAtWork:
related #907 

Some things to consider when updating:
* availability of these groups to take on additional work
* somethings dont need to be criteria for graduation - they can be things we've seen highly successful projects do and would strongly recommend others follow
* project Due Diligence is often project specific  - the current happenings and challenges in one project's process may look entirely different than another- we should plan for the common areas and provide guidance (perhaps) on domain or tech specifics.

xmulligan:
Slightly related, during Cilium's graduation process, we have also found that the document structure is different than in the repo, we based ours off of [Argo](https://docs.google.com/document/d/1R4WjMG9s9JX8onZvOzEFSjBBFAInurN8tSiAFLqj-FE/edit#heading=h.kd4eg2uz3lt0) and [Flux](https://docs.google.com/document/d/1Z6yPN9-bWeVGpMrBxXJ3NBTBYZowJ3R93wKHEVmBJ1A/edit#heading=h.1yxc6syuw6zj). 

The general structure is 

- Background
- Project Goal
- Current Status
- Project Progress since Incubation
- Future Plans
- Project Scope
- Value-add to the CNCF Ecosystem
- Alignment with other CNCF projects
- Formal Requirements
- Other Considerations

xmulligan:
Would it be helpful to create a DD template related to https://github.com/cncf/toc/blob/main/process/due-diligence-guidelines.md?

tomkerkhove:
As a maintainer going through a proposal now, that would definitely be useful and can be hosted on Google Docs thanks to https://github.com/cncf/toc/issues/980

rochaporto:
One more item to clarify on the graduation criteria:
```
[...]

PR #997, opened 2023-01-26
[WIP] What: Cloud Native Guide Posts Issue #961

TheFoxAtWork:
Why:

* Capture milestones from successful projects
* provide guidance and recommendations to projects seeking graduation
* Provide criteria to guide TOC on determining when sandbox projects should depart from sandbox

This change addresses the need by:

* Modifying archiving.md to reflect Sandbox Departure link
* Modify sandbox.md to set Sandbox Departure criteria
* Modify README.md to capture Milestones link
* create project milestones for community review, improvement, and feedback

Signed-off-by: Emily Fox <themoxiefoxatwork@gmail.com>

PR #925, opened 2022-09-14
Proposal for KEDA to become a CNCF Graduated project

tomkerkhove:
Over the past couple of years, we've had the pleasure of seeing KEDA grow from an open-source project that end-users started to play around with to an industry-leading app autoscaler.

Today, we are excited to open our proposal for KEDA to transition to a CNCF Graduated project and are happy to work with the CNCF TOC, TAGs, and community to evaluate our proposal.

Relates to https://github.com/kedacore/governance/issues/63

PR #1013, opened 2023-02-20
Clarify graduation submission requirements/review

duglin:
As I look at the graduation process for CloudEvents and KEDA it appears that the current process might need a bit of tweaking to help speed things up. Some observations/suggestions, from an outsider's point of view:
- not everything that is required of a project is documented. While there are some (what I would call) house-keeping requirements (like a COC, best practices badges), it's not clear what about the project and its community make it worthy of being "graduated". The graduation template does ask "why do you think you're ready?", but there's little in the way of guidance to indicate what the minimum bars are to to be approved. As a result, my impression is that it feels more like a "gut feel" of the TOC more than anything else. Granted, in the end that might be what it needs to be since having hard metrics to cover both "explosive growth projects" as well as "mature, slow moving but valued projects" might not be possible. However, we should document those cases where firm requirements are known (e.g. at least 2 orgs have maintainers), and then simply ask for the stats for the other metrics of interest. Either way, any data that the TOC uses should be documented as part of the graduation application. Very related to the discussion in https://github.com/cncf/toc/issues/992 .
- the TOC members are busy and therefore are the bottleneck when it comes to preparing the data that is to be reviewed. I'm proposing that we shift that burden to the project itself. To go along with the previous bullet, the graduation application should include all information the TOC members would like to review - then it's up to the project to provide that information. This shift the burden of the TOC (or the single TOC sponsor) from information gathers to application/information reviewer. If more information is needed then the TOC (or any other CNCF memebr) can ask for it via comments on the application PR.

To that end, this PR proposes a change to the graduation process:
- each graduation application becomes a PR to add the project's application to the github repo
- TOC/SIGs/TAGs/CNCF members... can then review and ask for additional information to be added if needed
- once all information is provided, TOC members simply LGTM the PR and when 6 or more have done so it can then be added to the next TOC call for a formal vote. Unless people are comfortable with the LGTM/async voting process - which would be ideal.

My ask, if this direction seems reasonable, is that the TOC members propose additional items to the application template in this PR so that it eventually includes all information they will need to make their decision. I believe that aside from speeding up the process, it will also increase the transparency of the the process.

/cc @tomkerkhove

### DIFF
--- a/process/graduation-proposal-template.md
+++ b/process/graduation-proposal-template.md
@@ -1,25 +1,199 @@
 # [Project] Graduation Proposal
 
-_**Insert introduction. See previous proposals for examples. This section should address from a broad perspective why the project feels they are ready to graduation and can state any major accomplishments or milestones.**_
+## Instructions:
+* replace all `???`s with the requested information.
+* elaborate on any prompt to explain the information provided.  Especially if
+  the data provided might seem inconsistent when compared to other projects.
+* for any prompt that requires a private answer put `available upon request`.
+* in most cases, the URLs provided should be to resources within the project's
+  GitHub repo or website rather than personal or private resources.
 
-## Graduation State Criteria
-_**Project should address each graduation criteria listed below**_
+## Project Summary
 
-### * Have committers from at least two organizations.
+1. Name of project: ???
+1. URL to project website:
+   * ???
+1. URL to GitHub repo:
+   * ???
+1. Project description (e.g. what it does, its value, origin, history):
+   * ???
+1. URL to list of project adopters (e.g. ADOPTERS.md or logos on the project 
+   website. For a spec have a list of adopters for the implementation(s) of the
+   spec):
+   * ???
+1. License type (should be Apache 2): ???
+1. URL to license:
+   * ???
+1. TOC sponsor name (if one exists): ???
+1. URL to Core Infrastructure Initiative
+   [Best Practices Badge](https://bestpractices.coreinfrastructure.org/):
+   * ???
+1. URLs to required reviews:
+   * Independent and 3rd party security audit results:
+     * ???
+1. URLs to governance documentation:
+   * ???
+1. Active maintainers:
+   * URL to document listing current active maintainers:
+     * ???
+   * Organizations with active maintainers as of the time of this PR (should
+     have at least 2 orgs):
+     * ???org (???#ofMaintainersFromOrg)
+1. URLs to previous CNCF applications (e.g. sandbox, incubator):
+   * sandbox: ???
+   * incubator: ???
 
-### * Have achieved and maintained a [Core Infrastructure Initiative Best Practices Badge](https://bestpractices.coreinfrastructure.org/).
 
-### * Have completed an independent and third party security audit with results published of similar scope and quality as [this example](https://github.com/envoyproxy/envoy#security-audit) which includes all critical vulnerabilities and all critical vulnerabilities need to be addressed before graduation.
+## Graduation Information
 
-### * Explicitly define a project governance and committer process. The committer process should cover the full committer lifecycle including onboarding and offboarding or emeritus criteria. This preferably is laid out in a GOVERNANCE.md file and references an OWNERS.md file showing the current and emeritus committers.
+1. Why is this project ready to graduate?
+   * ???
 
-### * Explicitly define the criteria, process and offboarding or emeritus conditions for project maintainers; or those who may interact with the CNCF on behalf of the project. The list of maintainers should be preferably be stored in a MAINTAINERS.md file and audited at a minimum of an annual cadence.
+1. What has changed since the project was approved for incubation?
+   * ???
 
-### * Have a public list of Project adopters for at least the primary repo (e.g., ADOPTERS.md or logos on the Project website). For a specification, have a list of adopters for the implementation(s) of the spec. Refer to [FAQs](https://github.com/cncf/toc/blob/main/FAQ.md#what-is-the-definition-of-an-adopter) for guidelines on identifying adopters.
+1. How can the project survive its founders and founding companies?
+   * ???
 
-## Incubation Details
-_**Project should address each area listed below**_
+1. Project adopters:
+   * Typical adopters (e.g. cloud providers, end users):
+     * ???
+   * Provide a list of adopters who are willing to chat with the TOC about
+     their experiences with the project and its community (about 2-4. Questions
+     might include: how/why do you use it? its stengths and weaknesses? 
+     are community interactions positive/responsive/timely?):
+     * ???name (???id at ???domainName)
+   * Adopter quotes with attribution:
+     * ???name : ???quote
 
-### * Link to Incubation Due Diligence(DD) Document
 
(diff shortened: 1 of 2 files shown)


process/graduation-proposal-template.md
```diff
@@ -1,25 +1,195 @@
...
+1. URL to Core Infrastructure Initiative
+   [Best Practices Badge](https://bestpractices.coreinfrastructure.org/):
+   * ???
+1. URLs to independent and 3rd party security audit results:
+   * ???
+1. URLs to governance documentation:
+   * ???
+1. Names of organizations that have current/active maintainers with the
+   maintainer GitHub IDs (should have at least 2 orgs):
+   * ???org (@???githubID, ...)
```
tomkerkhove, line 37: Or should we point to the maintainer overview of the given project to avoid stale information? In theory this should be well-documented for a graduation project

process/graduation-proposal-template.md
```diff
@@ -1,25 +1,195 @@
...
+1. URLs to previous CNCF applications (e.g. sandbox, incubator):
+   * sandbox: ???
+   * incubator: ???
 
-### * Have achieved and maintained a [Core Infrastructure Initiative Best Practices Badge](https://bestpractices.coreinfrastructure.org/).
 
-### * Have completed an independent and third party security audit with results published of similar scope and quality as [this example](https://github.com/envoyproxy/envoy#security-audit) which includes all critical vulnerabilities and all critical vulnerabilities need to be addressed before graduation.
+## Graduation Information
```
tomkerkhove, line 43: We should also add links to the required reviews, if any are mandatory

tomkerkhove (approved):
Great addition!

process/graduation-proposal-template.md
duglin: I modified the security audit bullet to make it so we can add more reviews under there as needed - then they're all in one spot. Are there other reviews I should add?

process/graduation-proposal-template.md
duglin: Good idea - I decided to do both... URL to maintainer doc AND the current list of orgs who have maintainers. I think having _today's_ # of orgs is a good bit of info for the TOC to have w/o having to follow a link to see it.

process/graduation-proposal-template.md
tomkerkhove: - TAG security was once required (but turned out external audit was enough) - https://github.com/cncf/toc/pull/925#issuecomment-1415993731
- TAG Contributor strategy was required for KEDA - https://github.com/cncf/toc/pull/925#issuecomment-1415993731

duglin: for TAG Contrib... let's see if we can create the list of requirements that the TAG used and then add them to this doc/PR. I've added a PR with some stuff from KEDA's review - see if anything is missing.

process/graduation-proposal-template.md
```diff
@@ -1,25 +1,212 @@
...
+   * Provide a list of adopters who are willing to chat with the TOC about
+     their experiences with the project and its community (about 2-4. Questions
+     might include: how/why do you use it? its stengths and weaknesses?
+     are community interactions positive/responsive/timely?):
+     * ???name (???id at ???domainName)
+   * Adopter quotes with attribution:
+     * ???name : ???quote
+
+
+## Due Diligence
```
craigbox, line 70: I disagree with the need for any due diligence at the graduation phase, unless there is any suggestion that a project may have regressed since entering the CNCF and may not continue to meet the incubation criteria.  That process could well happen with an incubating project and should be addressed by project review and/or the archival process, and need not be duplicated by graduation.

I take this position based on the expressed ideals that "[[the] incubation level is the point at which we expect to perform full due diligence on projects](https://github.com/cncf/toc/blob/main/process/graduation_criteria.md)", and that the stated intention of graduation is [a signal to the market that the software is ready for the early majority, not just early adopters](https://www.cncf.io/projects/#maturity-levels).

An exception would be that if a project intends to _join_ at Graduated, which has not yet happened but is explicitly mentioned as a possibility.  In that case, I would hope that the processes for "join at Incubation" and "demonstrate readiness for Graduation" were non-duplicative and could just both be followed.

duglin: I'm ok with removing it if the TOC would like that. I included it because:
1 - as you suggested, a project may want to join the CNCF at the graduated level
2 - since part of my goal is to reduce the amount of time for the TOC to review any pertinent information about a project, I wanted to put everything in one spot, even if that info was originally provided at a lower stage.
3 - worst case people could just copy the info from the info previously provided

I'll leave it to the TOC to decide if including this here is useful or not - I'll do whatever speeds up their graduation review process.

duglin:
I should point out that ideally I'd like to remove, not add, questions to the form that applicants need to fill out because it seems like many of the questions (while interesting) should not be used to determine if a project is "graduation worthy".  I say this because the answers are often transient or misleading. For example, "commit stats". For a younger project there could look really good, but once the project is very mature it may not look as good. But that doesn't mean the project isn't worthy... it could just mean it's stable. And in fact, too many commits could be a sign of being buggy. Plus, would a project lose it's "graduated status" if the answers to those questions suddenly changed? I doubt it. So, those questions shouldn't really be used to gauge graduation readiness.

And while I'm at it... there's part of me that even questions the need for the 3 levels at all since projects don't seem to be able to be demoted if things change (beyond being archived). Why should a once-good-but-now-not-so-good project remain at "graduated" status while newer projects that are wildly popular might be stuck at "incubator" stage? The status doesn't really provide the true value to the community who might view it as a way to gauge a project's worthiness. It's potentially very misleading and will only get worse as the entire Cloud Native community matures and project slowly fad in popularity.  As a result, I often wonder if there should only be one "bar"... either a project is a CNCF project or not. A project meets certain criteria (probably similar to what we have for incubator) and once they meet that then they're in. Very low bar and no more politics or games to worry about... just focus on the project being successful. But that might be a topic for a different day...

craigbox:
PTAL at #992 and #997. @duglin, it might be a good idea to attend the upcoming meeting if you can.

I think I'm in favour of your bar criteria, with 

* Associate CNCF Project

TheFoxAtWork:
@duglin we missed you at the task force presentation to the TOC today. Given your involvement with the group and the consensus the group achieved, do you believe this PR to be overcome by the group's deliverable/recommendations?  I realize the work is not "complete" in that the changes are not reflected in the TOC repo.

duglin:
As I'm not involved in the group any more I'll have to wait until the TOC repo changes are made before I can say for sure. The main things I'll be looking for are:
- near zero back-n-forth between project and CNCF. Ideally, a process like: project submits a PR and TOC votes.
- a single list of things that the project needs to do, or verify, to be approved - with results being part of the PR. Ideally a list in a single location and not spread out across multiple docs

TheFoxAtWork:
That is reasonable.  We'll leave this PR open, and I'll try to ensure those points you highlighted are covered or addressed by the resulting consensus reflected in the PR(s) to document those changes.

TheFoxAtWork:
Closing this PR as overcome by the recent changes to the TOC repo. Related issue to capture project feedback on this process: #1293

PR #1033, opened 2023-04-04
Update graduation_criteria.md

caniszczyk:
There have been some issues in CNCF where projects have waited too long to define basic governance processes, which have led to issues down the road as a project has scaled. It's better to encourage a project to start out with a simple governance structure in the beginning and evolve that over time versus starting with nothing and dealing with the aftermath of managing governance and growth at the same time.

There are also great resources and templates now that folks can lean on to craft their own governance https://contribute.cncf.io/maintainers/governance

### DIFF
--- a/process/graduation_criteria.md
+++ b/process/graduation_criteria.md
@@ -14,6 +14,7 @@ To be accepted in the sandbox a project must
 * Adopt the CNCF [Code of Conduct](https://github.com/cncf/foundation/blob/master/code-of-conduct.md)
 * Adhere to CNCF [IP Policy](https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy) (including trademark transferred)
 * List their sandbox status prominently on website/readme
+* Explicitly define a project governance and committer process. The committer process should cover the full committer lifecycle including onboarding and offboarding or emeritus criteria. This preferably is laid out in a GOVERNANCE.md file and references an OWNERS.md or MAINTAINERS.md file showing the current and emeritus committers. See https://contribute.cncf.io/maintainers/governance/ for more info.
 
 See the [CNCF Sandbox Guidelines v1.0](https://github.com/cncf/toc/blob/main/process/sandbox.md) for the detailed process.
 
@@ -24,8 +25,8 @@ See the [CNCF Sandbox Guidelines v1.0](https://github.com/cncf/toc/blob/main/pro
 To be accepted to incubating stage, a project must meet the sandbox stage requirements plus:
 
  * Document that it is being used successfully in production by at least three independent direct adopters which, in the TOC’s judgement, are of adequate quality and scope. For the definition of an adopter, see https://github.com/cncf/toc/blob/main/FAQ.md#what-is-the-definition-of-an-adopter.
-
  * Have a healthy number of committers. A committer is defined as someone with the commit bit; i.e., someone who can accept contributions to some or all of the project.
+ * Explicitly define the criteria, process and offboarding or emeritus conditions for project committers/maintainers; or those who may interact with the CNCF on behalf of the project. The list of maintainers should be preferably be stored in a MAINTAINERS.md file and audited at a minimum of an annual cadence.
  * Demonstrate a substantial ongoing flow of commits and merged contributions.
  * Since these metrics can vary significantly depending on the type, scope and size of a project, the TOC has final judgement over the level of activity that is adequate to meet these criteria
  * A clear versioning scheme.
@@ -39,7 +40,5 @@ To graduate from sandbox or incubating status, or for a new project to join as a
  * Have committers from at least two organizations.
  * Have achieved and maintained a Core Infrastructure Initiative [Best Practices Badge](https://bestpractices.coreinfrastructure.org/).
  * Have completed an independent and third party security audit with results published of similar scope and quality as the following example (including critical vulnerabilities addressed): https://github.com/envoyproxy/envoy#security-audit and all critical vulnerabilities need to be addressed before graduation.
- * Explicitly define a project governance and committer process. The committer process should cover the full committer lifecycle including onboarding and offboarding or emeritus criteria. This preferably is laid out in a GOVERNANCE.md file and references an OWNERS.md file showing the current and emeritus committers.
- * Explicitly define the criteria, process and offboarding or emeritus conditions for project maintainers; or those who may interact with the CNCF on behalf of the project. The list of maintainers should be preferably be stored in a MAINTAINERS.md file and audited at a minimum of an annual cadence.
  * Have a public list of project adopters for at least the primary repo (e.g., ADOPTERS.md or logos on the project website). For a specification, have a list of adopters for the implementation(s) of the spec.
  * Receive a supermajority vote from the TOC to move to graduation stage. Projects can attempt to move directly from sandbox to graduation, if they can demonstrate sufficient maturity. Projects can remain in an incubating state indefinitely, but they are normally expected to graduate within two years.


dzolotusky (approved):

dims:
💯 👍🏾

craigbox:
@jberkus FYI

tomkerkhove (approved):

process/graduation_criteria.md
```diff
@@ -14,6 +14,7 @@ To be accepted in the sandbox a project must
 * Adopt the CNCF [Code of Conduct](https://github.com/cncf/foundation/blob/master/code-of-conduct.md)
 * Adhere to CNCF [IP Policy](https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy) (including trademark transferred)
 * List their sandbox status prominently on website/readme
+* Explicitly define a project governance and committer process. The committer process should cover the full committer lifecycle including onboarding and offboarding or emeritus criteria. This preferably is laid out in a GOVERNANCE.md file and references an OWNERS.md or MAINTAINERS.md file showing the current and emeritus committers. See https://contribute.cncf.io/maintainers/governance/ for more info.
```
dims, line 17: I'd add "preferably machine parsable markdown", but it does not need to be here in this file, we can add that somewhere under https://contribute.cncf.io/maintainers/governance/

cathyhongzhang: We have these terminologies: committer, maintainer, owner, etc.  If there is no other place that clearly defines them, it is better to define them somewhere (could be in another place to which this file can reference)

TheFoxAtWork: @cathyhongzhang seems like our FAQs could use an update to include those but we should keep that separate for now as we've not had them articulated before and would want them to align with the language used on contribute.cncf.io and other LF materials.

cathyhongzhang: @TheFoxAtWork Agree. We need to articulate them clearly and be consistent across CNCF

TheFoxAtWork: I can reach out to TAG Contributor Strategy to see what they already have and figure out updates to our FAQs outside of this PR.  CC @jberkus @CathPag @geekygirldawn

aliok: We had a conversation around this with @TheFoxAtWork 

We (TAG Contributor Strategy) have a task for creating a less-technical glossary: https://github.com/cncf/tag-contributor-strategy/issues/140

craigbox: (On the topic of terminology, please have someone look at https://github.com/cncf/foundation/issues/433.)

process/graduation_criteria.md
```diff
@@ -24,8 +25,8 @@ See the [CNCF Sandbox Guidelines v1.0](https://github.com/cncf/toc/blob/main/pro
 To be accepted to incubating stage, a project must meet the sandbox stage requirements plus:
 
  * Document that it is being used successfully in production by at least three independent direct adopters which, in the TOC’s judgement, are of adequate quality and scope. For the definition of an adopter, see https://github.com/cncf/toc/blob/main/FAQ.md#what-is-the-definition-of-an-adopter.
-
  * Have a healthy number of committers. A committer is defined as someone with the commit bit; i.e., someone who can accept contributions to some or all of the project.
+ * Explicitly define the criteria, process and offboarding or emeritus conditions for project committers/maintainers; or those who may interact with the CNCF on behalf of the project. The list of maintainers should be preferably be stored in a MAINTAINERS.md file and audited at a minimum of an annual cadence.
```
RichiH, line 29: Is there a good reason not to upgrade the existence and review of MAINTAINERS.md from "should" to "must"?

craigbox: Please don't apply "must" to the file format. Istio keeps its list in yaml format as it is an input to [peribolos](
https://docs.prow.k8s.io/docs/components/cli-tools/peribolos/).

process/graduation_criteria.md
lizrice: It makes sense for CNCF as a body to have a consistent understanding of what is meant by these different terms, but I don't think it makes sense to impose specific terms on incoming projects, given that no particular governance model is required. An established, successful project shouldn't have make-work imposed on them to rename any roles and terms they may already have in their governance, if that governance is working successfully for the project

TheFoxAtWork: There are two issues here: 
1) defining the roles so they make sense in the context of governance as a criteria for moving levels (note this specific line is about *sandbox projects* and would impact the TOC's ability to accept them if this is NOT met) and
2) the roles that can act on and receive CNCF benefits (voting, services, etc.)

Given the TOC principles [allow a project to self-govern](https://github.com/cncf/toc/blob/main/PRINCIPLES.md#projects-are-self-governing), the specifics of HOW they choose to define a project governance and committer/maintainer/SteeringCommittee/other process is up to them to do, so long as it exists. Given this specific line is around sandbox, I believe the 'should' statement is acceptable. Projects that apply for sandbox are well positioned if they have more than two or three maintainers, but more-often-than-not are not thinking about their longevity or success until something happens that requires it (thus this being introduced early to think about and consider).

From the principles: 
> governance model that includes a **contribution-based process by which contributors can become committers or maintainers**, and a **clear definition of the top-level project leadership**, with which the foundation will engage and inform and from which it may receive requests for funding and support. 

Separating this discussion out: 
1) Is the language within the changes proposed sufficient to move forward with the expectation that it will change with better definition and expectation of those terms (to be done [outside of this PR](https://github.com/cncf/tag-contributor-strategy/issues/140))?  or does the TAG CS issue need completed first?

2) recommend this be moved to [Craig's cited issue](https://github.com/cncf/foundation/issues/433) 
Those terms carry weight in how the project receives CNCF benefits and as Liz states we shouldn't impose additional work to restructure project governance to conform with that - it goes against the principle I mentioned. If a project has its own roles defined, the project just needs to state to the CNCF of those roles which can act on those benefits (top-level project leadership). 

For myself (reviewing sandbox applications, due diligence, or other project status and health activities) - so long as I can find their roles and understand them, I can align that with the governance needs for moving levels and their ability to exercise those responsibilities in the operation of the project. If I can't get that alignment from what is written, I seek clarity with the project until that alignment is readily conveyed within the governance that is documented.

jberkus: Can we please stop using the word "committer"?  It is not immediately comprehensible to modern projects, and it's different from the word we use for project leadership elsewhere, which is "maintainer".

jberkus, line 17:
```suggestion
* Explicitly define a project governance and maintainer selection process. The maintainer process should cover the full maintainer lifecycle including onboarding and offboarding or emeritus criteria. This preferably is laid out in a GOVERNANCE.md file and references an OWNERS.md or MAINTAINERS.md file showing the current and emeritus maintainer. See https://contribute.cncf.io/maintainers/governance/ for more info.
```

Please stop using the word "committer".  It is misleading and confusing to projects.

jberkus, line 17: I'm also going to suggest that an OWNERS file is not a substitute for MAINTAINERS.md.  This is because the owners file contains ONLY github handles, and not contact information, affiliation, full name, or any other information about the maintainer.  If the goal is to make the project more transparent to the TOC and potential contributors, then we need a full maintainers file.

process/graduation_criteria.md
jberkus, line 29:
```suggestion
 * Explicitly define the criteria, process and offboarding or emeritus conditions for project maintainers; or those who may interact with the CNCF on behalf of the project. The list of maintainers should be preferably be stored in a MAINTAINERS.md file and audited at a minimum of an annual cadence.
```

Still removing the word "committers".

process/graduation_criteria.md
```diff
@@ -39,7 +40,5 @@ To graduate from sandbox or incubating status, or for a new project to join as a
  * Have committers from at least two organizations.
  * Have achieved and maintained a Core Infrastructure Initiative [Best Practices Badge](https://bestpractices.coreinfrastructure.org/).
  * Have completed an independent and third party security audit with results published of similar scope and quality as the following example (including critical vulnerabilities addressed): https://github.com/envoyproxy/envoy#security-audit and all critical vulnerabilities need to be addressed before graduation.
- * Explicitly define a project governance and committer process. The committer process should cover the full committer lifecycle including onboarding and offboarding or emeritus criteria. This preferably is laid out in a GOVERNANCE.md file and references an OWNERS.md file showing the current and emeritus committers.
```
jberkus, line 42: OK, we don't want to just remove this without replacing it.  By the time a project has gotten to graduation, they should have evolved (and, for that matter, used) their governance.  That is, the expectations for transparency, accessibility, and completeness for a graduated project should be higher than they were for sandbox, and we should call that out as a requirement.

Draft text:

Have complete and detailed governance, both in documentation and in practice, that address all aspects of managing the project, including contributor roles, maintainer selection and departure, handling code of conduct and security incidents, and management of any subprojects.  This governance should be open to all contributors and non-prejudicial towards any contributor affiliation.

process/graduation_criteria.md
craigbox: > it's different from the word we use for project leadership elsewhere, which is "maintainer".

Not all committers are project leadership, especially in larger projects.  (See [Craig's cited issue](https://github.com/cncf/foundation/issues/433).)

Is the intention of this criteria to only refer to project leadership?  I read it as talking about other kinds of project membership too.

jberkus (changes requested):

jberkus:
We should also suggest specific templates for "minimal governance".  Would it be sufficient to simply link to the [Maintainer Council template](https://contribute.cncf.io/maintainers/templates/governance-maintainer/)?  Or do we want something even simpler than that?

process/graduation_criteria.md
mrbobbytables: Commiter is the term used in the foundation charter. A small preamble in the doc like 'For the purposes of the TOC docs and process, a maintainer is equivalent to a committer as defined by the CNCF charter' or something like that.

jberkus: Additionally, I had the impression that this change WAS intended only to define how project leadership is selected.  @TheFoxAtWork, can you clarify?

jberkus: Although: the combination of a README with the extra information together with an OWNERS file works.  It might be better to define what information needs to be exposed about maintainers.

process/graduation_criteria.md
jberkus: Also, question: if we're requiring governance at sandbox entry, what are we requiring at Incubation?

process/graduation_criteria.md
craigbox: Could I kindly ask that instead of the TOC defining apples to mean oranges, that any change is holistic and updates the term used in the CNCF charter as well?

lizrice: > Again, few current github-based projects use the term "committer",

Cilium is an example of a project that uses this term. And like @mrbobbytables said, it's in the charter. 

Before going to the effort of changing the charter, maybe step back and see what problem this is trying to solve? Given that we don't impose any specific governance, and projects likely come into the organization with some pre-defined governance docs in place, forcing everyone into the same terminology is just make-work. It doesn't **help projects**.

lizrice: It *is* currently defined: it needs to show who the current and emeritus committers are. We don't need to determine exactly what is and isn't required or how the information is presented - when assessing a project, the TOC can simply look at the file and consider whether information about committers is present

This drive towards consistency across projects feels like a slippery slope towards yet more make-work process requiring busy project maintainers to re-write files to fit some arbitrary format. How does consistency between projects really help people who want to contribute to a project? Most folks concentrate on one thing at a time.

aliok: TAG CS now has the first version of the non-technical glossary: https://contribute.cncf.io/resources/glossary/#project-maintainer.

craigbox: > In CNCF, term “maintainer” is used over other similar terms such as “committer”.

But not necessarily in CNCF _projects_; I don't think it is the intention of the TOC to make projects use a standard noun, or if it is to be part of the 'light' oversight that projects are subject to, that it be very clearly communicated.

aliok: Thanks for the feedback @craigbox 
I would be very happy to update the non-technical glossary with the outcome of this discussion (or a separate discussion).

process/graduation_criteria.md
lizrice: See above, we use the word "committers" in the charter so it seems strange to erase it here

process/graduation_criteria.md
lizrice: Raised an issue against the [Glossary](https://github.com/cncf/tag-contributor-strategy/issues/471) about this.

lizrice: I will echo what @craigbox says here: when projects sign up to join the CNCF they are assured that they won't be required to adopt a standard governance model. As a consequence, they have no expectation that they will be required to adopt some standardized language within their existing governance documentation

lizrice:
While it's great to have sample governance models and templates for projects to look at, there are projects that join the CNCF as established, successful, open, widely-adopted projects, with existing governance models that have served them well for years. We promise them that we will treat them with a light touch, so let's make sure that in trying to provide guidance to new and less-established projects, we don't turn that into additional requirements and extra work for these maintainers.

process/graduation_criteria.md
jberkus: Yeah, I'm trying to scrub it everywhere I see it.  And we should amend the charter to remove it.

craigbox: I defer to your experience here Josh, but I don't understand the motivation - could you share more please? Is this a personal request, or to align with some broader industry change I was not aware of?

cathyhongzhang (approved):

cathyhongzhang:
As a result of today's TOC meeting, let's hold off on merging this PR.  The recommendations from the task force will likely alter the criteria from what is recommended here. Once the recommendations from the task force are finalized and public comment is complete, we can determine if this PR is overcome by those recommendations

process/graduation_criteria.md
jberkus: Because, if you use git, *every single contributor is a committer*.  Look at the "committer" stats in DevStats, the official metrics for CNCF projects.  They include *everyone with a merged PR*.

So, for any potential contributor who's been in OSS less than 10 years, using the word "committer" to refer to project leadership is profoundly confusing.  It's the equivalent of describing a retail business, and using the word "associates" as being the same as "management". So, yes, it's an industry change.  Every time a TOC member says "committer" they sound 60 years old.

In addition to this, many/most CNCF projects have people who review and approve PRs who are not otherwise project leadership -- and people who are project leadership who don't do much approving PRs.   

And, further, all of the documentation from CNCF staff, including all of the mailing lists, etc., refer to "maintainers".  Having a single, clear term, which doesn't mean two radically different things is much better for everyone.

lizrice: > Every time a TOC member says "committer" they sound 60 years old.

This ageist comment has no place in a CNCF discussion.

It feels to me like you've made this point about changing "committer" to "maintainer" over and over, but there's no evidence of a groundswell of support. I will keep making the point that we do not define governance, so projects can use whatever terminology suits them best. The language of the Charter uses the word "committer", and changing this needs a Governing Board discussion, so unless someone from the TOC and/or GB want to take it up, going round in circles on this doesn't seem like a great use of time.

raravena80: I echo Liz's sentiment about the 60 year old comment here.

The comment could imply that there is a requirement for a TOC member to be under 60 years old.

TheFoxAtWork:
Closing this PR as this consideration was added into the recent matriculation process changes and recommendations from the task force.

PR #1052, opened 2023-05-17
Organize reviews folder 

jeefy:
As we continue collecting reviews, the root `reviews` folder is ballooning out a bit. Let's reign it in with some better structure, that will also aid in our future automation. :)


amye (approved):

TheFoxAtWork (commented):
While i love the idea of this restructure, there is some inconsistencies between the levels folder structure.  Graduation and incubation don't have years (not sure they should), but sandbox does (also not sure it should).  So long as the files are dated accordingly, an individual that is looking at those files is likely doing so for a specific project and if needing to reference the year can find that within the document. However if we leverage the year in automation, reporting or other construct, then the folder structure should reflect the year (if that is how such automation is intended to work)

PR #1060, opened 2023-05-22
Fix whitespace and newlines

RichiH:
It's a nit PR, but as I was just using CLI to make the edits, all of those stood out


stmcginnis (approved):
Nice cleanup.

PR #1063, opened 2023-05-29
Update PRINCIPLES.md

amye:
This updates PRINCIPLES with Leadership Principles from TOC.
A minimum of two weeks public comment period opens today, May 29th with a TOC vote to follow if we feel we've resolved any feedback.

PR #1065, opened 2023-05-31
Update PRINCIPLES.md

joannalee333:
Adding "ideally" to having a neutral third party providing feedback since there may be times when a third party is not available so I suggest making this flexible


dims:
+1

TheFoxAtWork (approved):

amye:
Resolved in https://github.com/cncf/toc/pull/1063/commits

PR #1068, opened 2023-06-02
OpenYurt 2022 Annual Review

gnunu:


cathyhongzhang (approved):
The project is progressing well and has a healthy community. There are other projects that extend the cloud to the edge, such as KubeEdge. It will be good to describe the difference in terms of functionality or design.

PR #1081, opened 2023-06-13
General process steps

TheFoxAtWork:
What: Pull in the general steps for moving levels
Why:

* Some projects shared it was not clear what was next or who was responsible for performing it.

This change addresses the need by:

* adding General steps section
* documenting general steps
* document the "who" for general steps

### DIFF
--- a/process/project_proposals.md
+++ b/process/project_proposals.md
@@ -70,12 +70,14 @@ While the details of the process are described in detail further for Incubating 
   * Who: TOC Sponsor(s) and Project
 * Assuming all outstanding issues are resolved, the TOC opens an internal comment period, about 1 week, for other TOC members to perform an independent review and verify all areas of the project have been evaluated
   * Who: TOC Sponsor(s) and TOC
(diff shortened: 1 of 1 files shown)


tomkerkhove:
> * Provided not additional items are identified during the public comment period, the TOC opens voting on the mailing list. The project is updated on the project board to 'In Voting'.
>  * Who: Initiate - CNCF Support Staff for the TOC, Voting - TOC and community members

I think it would be nice to know what expected timeline is used here? Being a KEDA maintainer, we'll soon be opening TOC voting and we don't know how long that is expected to take. If this is unknown/"it depends", then let's just define this here as well.

> * If the vote passes, the results are emailed and the project is placed in the 'Done' state on the project board.

I believe there is a step after this: Announcement. I think we need to add the following:

> Once the process completed with positive outcome, the result gets announced on the CNCF blog and press release gets created
>   * Who: CNCF Staff

At least this has been the case for KEDA after Sandbox + Incubation.

nikhita (approved):
LGTM

PR #1083, opened 2023-06-13
add initial gitvote config

jeefy:
This adds a git vote config!
`/vote` will be a 26w timeboxed 66% vote
`/vote-1wk` will be a 1w timeboxed 66% vote

### DIFF
--- a/.gitvote.yml
+++ b/.gitvote.yml
@@ -1,0 +1,15 @@
+profiles:
+  default:
+    duration: 1w
+    pass_threshold: 66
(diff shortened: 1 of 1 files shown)


amye:
Let's flip it to just /vote because we'll use it for all kinds of things, not just incubation and use /vote-1week for one week votes.

jeefy:
Updated config (and initial PR desc)

amye:
Ready to rock :D

PR #1107, opened 2023-07-03
proposal: Automatically post voting results every week

tomkerkhove:
Automatically post voting results every week so that community can track progress without having to post the command.

Relates to https://github.com/cncf/gitvote/issues/162


craigbox:
Why not every day?

I would hope that by the point a vote is called, enough information should have been presented (and enough consensus gathered) that members should know which way they are going to vote, and thus the expectation should be that it happens within a much shorter period.

PR #1150, opened 2023-08-17
Remove projects.md

castrojo:
We've added the security and fuzzing information to fields in the landscape metadata, so you don't need to track this by hand here any more. Data is there and landed, just waiting on the frontend so I'll mark this as draft until then.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205339026797896


castrojo:
Ok just a quick update, we've got the security and fuzzing audits supported in the Landscape now, here's a screenshot:

image

We've also added support for the projects to have a direct URL to their TAG's github repo so it's more clear which TAGs relate to which projects. That's not a hyperlink yet but work in progress. 

That should cover everything we need to not have to manually track this in a separate sheet.

castrojo:
I'd like to PR in instructions so that people understand that they can add the fuzzing and security audits to the landscape. 

However I am finding that the security reports are done at the project level, I can't seem to find a centralized document that outlines the steps for doing an audit. (Perhaps in another repo?)

Basically I want to add something like "[ ] Submit your audit/fuzzing audit to the Landscape (example link)" as part of the process.

caniszczyk:
We don't have a place to do that @castrojo as it's handled via servicedesk atm. We can include it when we work with vendors in the contract.

PR #1183, opened 2023-09-27
first pass at stalebot config

jeefy:
This will stale-then-close any `wontfix` PR (we typically should leave PRs open in this repo)

This will stale-then-close any issue not labeled with `enhancement` or `project onboarding` (those seemed the two best exemptions)

Timing is: Mark stale after 60d, Auto-close 14d after


amye:
LGTM, let's try it.

PR #1185, opened 2023-10-06
Add LFX Insights as an onboarding task for the CNCF in the issue template

krook:
As projects go through sandbox onboarding, we'd like to add them to Insights as well as devstats. This line should also be added to the existing issues that predate this change to the template. I can work with you on that @Cmierly.

Also, one commit in here fixes an issue with the Markdown for the Annual Report template. Other very minor fixes are from linting the Markdown before committing.


lukaszgryglicki:
I'm always adding new projects to DevStats, once they are added.
There is also a tool that checks for any differences in DevStats and landscape projects list and reports difference so I can add or update DevStats (every day).

PR #1188, opened 2023-10-12
Update PROJECTS.csv

craigbox:
Some level updates.  I don't think this is state of the art anymore but it's very useful and I wish it were still maintained.  Can it be generated from the current source of truth?


amye:
Landscape.cncf.io is the source of truth, we should deprecate this properly instead of accidentally

craigbox:
see https://github.com/cncf/landscape/issues/3555 for why I asked

Riaankl:
This file would be removed with #1309 and it's related file was removed by #1150.

Both these files are no longer updated and replaced by Landscape2.

PR #1216, opened 2023-12-07
Add new project level templates

castrojo:
This is a templated version of the items discussed from the project-moving-levels task force. It's split into three documents, one for each level. I added a number so they display in order but ideally there's a github issue template that will just link to these directly. 

The idea is a project files a tracking issue, and then PRs in one of these and then that becomes the living document of the lifecycle of the project for that level. There are placeholders for the TOC to be able to fill in the evaluation for each criteria bullet. Happy to make adjustments and fixes depending on feedback!

### DIFF
--- a/proposals/0 - sandbox-template.md
+++ b/proposals/0 - sandbox-template.md
@@ -1,0 +1,176 @@
+---
+name: Sandbox Due Diligence
+about: Projects Moving Levels
+title: "[Sandbox DD] project"
+labels: sandbox
+assignees: amye, castrojo
+--- 
+
+# Information Summary
+
+- Link to [Sandbox submission issue]()
+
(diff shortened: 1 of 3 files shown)


TheFoxAtWork:
I'd like to shift the flow a bit here. I think the three templates is good, and i like the project filing the issue to move levels, but i want the TOC to be the author of the PR since it should be an evaluation by the TOC.  this will enable us to link it to the issue so the project receives the notification and the project can provide comments/clarity/questions on the evaluation by the TOC.

TheFoxAtWork:
@castrojo thank you for taking the time to put this together, the changes you've recommended with some additions are reflected in the TOC repo as result of the moving levels task force recommendations. This PR is overcome by that event.

PR #1305, opened 2024-04-18
Update TAG app delivery personell

lianmakesthings:
- Alois stepped down as Chair
- Lian was voted in as Chair and confirmed by TOC
- Alex Jones stepped down as Tech Lead
- updated some other information in various places where they were missed

@Riaankl

### DIFF
--- a/tags/cncf-tags.md
+++ b/tags/cncf-tags.md
@@ -31,7 +31,6 @@ TOC and TOC Contributors have fulfilled TAG duties in the past and will continue
 * [Thomas Schuetz](https://github.com/thschue)
 
 #### Tech Leads
-* [Alex Jones](https://github.com/alexsjones)
 * [Karena Angell](https://github.com/angellk)
 * [Lian Li](https://github.com/lianmakesthings)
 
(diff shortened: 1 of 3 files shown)


Riaankl:
@lianmakesthings I would propose inline with the new TAG Charter template that I shared the TAG App delivery to remove the leadership information from the charter, instead of updating it.

/CC @thschue

angellk (changes requested):
Please add Alois to Technical Leads

Riaankl:
> @lianmakesthings I would propose inline with the new TAG Charter template that I shared the TAG App delivery to remove the leadership information from the charter, instead of updating it.
> 
> /CC @thschue

@angellk do you agree that the  leadership should be removed from the charter in line with the new template?

TheFoxAtWork (approved):

kevin-wangzefeng:
+1 to  "the leadership should be removed from the charter".
But I don't think it's a blocking issue of this PR, as there are other TAG charters content leadership info already, we  can fix them with a cleanup PR after the new TAG Charter template ready.

riaankleinhans:
@kevin-wangzefeng I agree and merged. Thanks

The Charter & README Temples is ready and can found [here](https://docs.google.com/document/d/1ut60KmkIwggpnF0Wtb8-qoECnB9W0qq8nihqKxYQ_a0/edit#heading=h.4zf0p5wp9r8t)

PR #1309, opened 2024-04-30
Remove projects.csv as it is replaced by landscape2

riaankleinhans:
The function of this file was replaced by Landscape2 and project update should be added to the [landscape/landscape.yml](https://github.com/cncf/landscape/blob/master/landscape.yml) file.

The related file was already removed by #1150


craigbox:
Is there anywhere else that the sponsors of the projects are tracked, or is this deemed important?  That is being lost by this change.
(It looks like all the other relevant data is now in the landscape.)

TheFoxAtWork:
The sponsors are usually tracked in the DD document and typically not important once the project has moved levels and often the TOC sponsor has moved on so identifying who they are often wont provide value on the circumstance we would look for that information.

TheFoxAtWork (approved):

Issue #1303, opened 2024-04-16
[SANDBOX PROJECT ONBOARDING] Radius


jeefy:
Welcome to CNCF Project Onboarding!

This is an issue created to help onboard your project into the CNCF after the TOC has voted to accept your project into the Sandbox.

We would like to complete onboarding within one month of acceptance.

**From the project side, please ensure that:**

- [ ] You understand the project proposal process and requirements: <https://github.com/cncf/toc/blob/main/process/project_proposals.md#introduction>
- [ ] You understand the services available for your project at CNCF <https://www.cncf.io/services-for-projects/>
- [ ] You ensure your project meets the CNCF IP Policy: <https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy>
- [ ] You review the online programs guidelines: <https://github.com/cncf/foundation/blob/master/online-programs-guidelines.md>
- [ ] You understand the trademark guidelines: <https://www.linuxfoundation.org/en/trademark-usage/>
- [ ] You understand the license allowlist: <https://github.com/cncf/foundation/blob/master/allowed-third-party-license-policy.md#approved-licenses-for-allowlist>
- [ ] Your project is working on written, open governance. See <https://contribute.cncf.io/maintainers/governance/>
- [ ] Your Slack channels are migrated to the Kubernetes or CNCF Slack workspace. See <https://slack.com/help/articles/217872578-Import-data-from-one-Slack-workspace-to-another> for more details)
- [ ] Your project in its own separate neutral GitHub organization. This will make it transferable to the CNCF's GitHub Enterprise account.
- [ ] You submit a pull request to add your project as a Sandbox project to <https://landscape.cncf.io>
- [ ] You create maintainer list + add to aggregated <https://maintainers.cncf.io> list by submitting a PR to it
- [ ] You submit a pull request to <https://github.com/cncf/artwork> with your artwork
- [ ] You transfer your domain to the CNCF - <https://jira.linuxfoundation.org/plugins/servlet/theme/portal/2/create/63>

**Things that CNCF will need from the project:**

- [ ] Provide emails for the maintainers added to <https://maintainers.cncf.io> in order to get access to the maintainers mailing list and Service Desk - <project-onboarding@cncf.io> is the best email to send those to
- [ ] Transfer any trademark and logo assets over to the LF - <https://github.com/cncf/foundation/tree/master/agreements> has agreements
[...]

PR #1319, opened 2024-05-10
Clarifying edits for new project onboarding.

bridgetkromhout:
Results of discussion with @jeefy about ambiguities in the current template. Thanks to @jonvsm for the inspiration in https://github.com/cncf/toc/issues/1303#issuecomment-2103378500.

### DIFF
--- a/.github/ISSUE_TEMPLATE/project-onboarding.md
+++ b/.github/ISSUE_TEMPLATE/project-onboarding.md
@@ -10,7 +10,9 @@ Welcome to CNCF Project Onboarding!
 
 This is an issue created to help onboard your project into the CNCF after the TOC has voted to accept your project into the Sandbox.
 
-We would like to complete onboarding within one month of acceptance.
+We would like your project to complete onboarding within one month of acceptance.
+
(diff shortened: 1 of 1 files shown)


TheFoxAtWork (approved):
not necessarily a request for change, the way we discuss slack migration implies (to me) that there is an expectation that project specific slacks dont exist, when they do.  I dont know if projects have this confusion but wanted to mention it here on the chance it comes up in the future. Thank you @bridgetkromhout

bridgetkromhout:
> not necessarily a request for change, the way we discuss slack migration implies (to me) that there is an expectation that project specific slacks dont exist, when they do. I dont know if projects have this confusion but wanted to mention it here on the chance it comes up in the future. Thank you @bridgetkromhout

Reasonable point, @TheFoxAtWork. Perhaps the nuance we're looking for is that we _can_ migrate Slack channels between teams (which people might be unaware is possible), and if you want a migration, cool (possibly _encouraged_ since then CNCF can handle code of conduct stuff there better); also, I added the bit about getting the relevant Slack channels newly created if you don't start with pre-existing channels, because that was also confusing to this Discord-using project.

PR #1324, opened 2024-05-13
Add README and charter templates

riaankleinhans:
Add templates for TAGs to use where rewriting the Charter and README. 
This folder contain several place holder template documents that still need to be created, therefore this seen to best place for these templates.


TheFoxAtWork (approved):

linsun (commented):
LGTM

@riaankleinhans @TheFoxAtWork Thanks for pulling this together. One thing we may want to consider to add as a follow up is the process to manage lifecycle of the WG (creation and remove inactive WG).

dims:
LGTM (+1 to iterate!)

PR #1341, opened 2024-05-28
Update TAG App Delivery and TAG Network in cncf-tags.md

angellk:
Update to reflect current TAG leadership


riaankleinhans:
Thank you @angellk for the update.
Please also add Lin & Richard to the [emeritus_leaders.md](https://github.com/cncf/toc/blob/main/tags/emeritus_leaders.md) file.

PR #1355, opened 2024-06-19
Add vote announcements to TOC votes

riaankleinhans:
Adding the new [Git Vote Announcements Function](https://github.com/cncf/gitvote?tab=readme-ov-file#announcements) to the TOC votes.

This would result in an announcement in the TOC Git Hub Discussions under [Announcements](https://github.com/cncf/toc/discussions/categories/announcements) as soon as a vote is concluded. The Announcements Announcement Discussion tag not used for anything else at this time, so it should be a good fit.

We might want to "market" this outcome so people know to subscribe to the Announcements channel.

### DIFF
--- a/.gitvote.yml
+++ b/.gitvote.yml
@@ -16,3 +16,7 @@ profiles:
         - cncf-toc
       exclude_team_maintainers: true
     close_on_passing: true
+  announcements:
+     discussions:
+       category: announcements 
+       
--- a/.gitvote.yml
+++ b/.gitvote.yml
@@ -8,6 +8,9 @@ profiles:
         - cncf-toc
       exclude_team_maintainers: true
     close_on_passing: true
+    announcements:
+      discussions:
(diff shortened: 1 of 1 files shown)


.gitvote.yml
```diff
@@ -16,3 +16,6 @@ profiles:
         - cncf-toc
       exclude_team_maintainers: true
     close_on_passing: true
+  announcements:
```
tegioz, line 19: The announcements section should be inside one of the profiles @riaankleinhans (`default` or `1wk` in this case). This allows you to enable this feature only in some profiles, or using different categories for announcements on each profile when needed 🙂

tegioz (changes requested):

.gitvote.yml
riaankleinhans: Update, thank you for the review.

tegioz (approved):

PR #1440, opened 2024-09-18
Reorg of proposals and reviews

mrbobbytables:
This PR switches the structure from organizing reviews/proposals for each level to a directory per project. It should make artifacts easier to find and supporting material for can be added by the projects into their own directories.


dims:
looks like the wasmedge one got doubled up and we dropped the correct one. 

LGTM. thanks @mrbobbytables

mrbobbytables:
@dims looks liks they were both named 2022 review, but one was actually 2023. I've resolved it now :+1:

angellk (approved):
lgtm - thanks for re-organizing @mrbobbytables !

PR #1500, opened 2024-12-06
Update runtime-charter.md

raravena80:
Up to date info


angellk (approved):
Thanks for updating with the current TAG Chairs @raravena80

PR #1501, opened 2024-12-07
Fix Code of Conduct requirement.

jberkus:
Both the incubating and graduating templates have this requirement:

- [ ] **Document agreement that project will adopt CNCF Code of Conduct.**

This doesn't make any sense, since a CNCF project is required to adopt the CNCF CoC by the end of onboarding.  They don't get to wait until Graduation.  As such, I've replaced it with this requirement:

- [ ] **Document adoption of the CNCF Code of Conduct**


angellk (approved):
Thanks @jberkus - good catch.

Issue #1484, opened 2024-11-14
[Incubation] kubegateway Incubation Application


ilevine:
# K8sGateway Incubation Application
v1.6
This template provides the project with a framework to inform the TOC of their conformance to the Incubation Level Criteria. 

Project Repo(s): https://github.com/k8sgateway/k8sgateway, https://github.com/k8sgateway/community

Project Site:  https://k8sgateway.io

Sub-Projects: NA

Communication: https://cloud-native.slack.com/archives/C080D3PJMS4

Project points of contacts: Idit Levine, Lin Sun, Yuval Kohavi

- [ ] (Post Incubation only) [Book a meeting with CNCF staff](http://project-meetings.cncf.io) to understand project benefits and event resources. 

### Application Level Assertion

- [ ] This project is currently Sandbox, accepted on YYYYMMDD, and applying to Incubation.
- [X] This project is applying to join the CNCF at the Incubation level.

### Adoption Assertion

_The project has been adopted by the following organizations in a testing and integration or production capacity:_
- https://k8sgateway.io/#users 

### Suggested

N/A

### Required

- [ ] **Engage with the domain specific TAG(s) to increase awareness through a presentation or completing a General Technical Review.**
  - This was completed and occurred on DD-MMM-YYYY, and can be discovered at $LINK.

 

- [ ] **TAG provides insight/recommendation of the project in the context of the landscape**

 

- [ ] **All project metadata and resources are [vendor-neutral](https://contribute.cncf.io/maintainers/community/vendor-neutrality/).**

 

- [ ] **Review and acknowledgement of expectations for [Sandbox](https://sandbox.cncf.io) projects and requirements for moving forward through the CNCF Maturity levels.**		
- Met during Project's application on DD-MMM-YYYY.

 

- [ ] **Due Diligence Review.**

Completion of this due diligence document, resolution of concerns raised, and presented for public comment satisfies the Due Diligence Review criteria.

- [ ] **Additional documentation as appropriate for project type, e.g.: installation documentation, end user documentation, reference implementation and/or code samples.**

 

## Governance and Maintainers

Note: this section may be augmented by the completion of a Governance Review from TAG Contributor Strategy.

### Suggested

- [ ] **Clear and discoverable project governance documentation.**

 

- [ ] **Governance has continuously been iterated upon by the project as a result of their experience applying it, with the governance history demonstrating evolution of maturity alongside the project's maturity evolution.**

 

[...]

PR #1514, opened 2025-01-06
Naming convention guidance for new projects

dims:
We have an incoming request from Kubernetes Steering about names for new incoming projects:
https://lists.cncf.io/g/cncf-toc/message/8726

If this PR has consensus and is merged, then follow ups would be:
- Request k8s steering to publish their naming guidelines in a well publicized location (or linked from a well somewhere easily findable)
- Check if other popular projects want to do the same? (otel? envoy?)
- May be even have a central location that point to all the individual project naming conventions
- If a project has its own donation process for sub-projects, they should be linked from the project specific naming guidelines as well.

NOTE: This is not a retro-active policy, we don't go back to revisit projects that already use popular prefixes to change them, this is meant to be forward looking guidance for projects that will be processed by TOC once this merges.

### DIFF
--- a/process/README.md
+++ b/process/README.md
@@ -23,6 +23,10 @@ Projects may enter the CNCF either by applying as a sandbox project or applying 
 
 Evaluting projects against the criteria does take some time and the TOC has recently modified the Due Diligence process to reduce duplication of work, streamline handoffs, and provide better transparency to the project and its adopters about its conformance and implementation of the criteria.
 
+##  Naming conventions for projects
+
+All CNCF projects are subject to the [Trademark Usage Policies](https://www.linuxfoundation.org/legal/trademark-usage) set by Linux Foundation. Specifically, new or incoming projects should avoid use existing trademarks in their proposed project names. In addition, if they are intending to use a popular prefix/suffix of an existing project (like "kube" or "k8s"), then they should consult the leadership group of the respective project to seek their approval and document the consensus reached. Existing projects are encouraged to document their naming guidelines to make this process smooth as well to avoid lengthy deliberation process for new project names.
+
 ### Project resources and guide posts
 
 The TOC, with support from the [Technical Advisory Groups](/tags/README.md), have a wide variety of resources available to assist projects.  Current and aspiring maintainers of cloud native projects can find a lot of information and templates on [contribute.cncf.io/maintainers](https://contribute.cncf.io/maintainers/). The TOC also maintains [project Guide Posts](../docs/project_guideposts.md) - a collection of guiding points that have assisted cloud native projects as they grow and mature in the ecosystem. These are not requirements for moving levels, those may be found in the respective application issue templates ([Incubation](../.github/ISSUE_TEMPLATE/template-incubation-application.md), [Graduation](../.github/ISSUE_TEMPLATE/template-graduation-application.md)). The [Guide Posts](../docs/project_guideposts.md) are a resources for projects to leverage that is beneficial in meeting or exceeding the criteria defined.


dims:
Note sent to k8s steering in https://github.com/cncf/toc/pull/1514

process/README.md
```diff
@@ -23,6 +23,10 @@ Projects may enter the CNCF either by applying as a sandbox project or applying
 
 Evaluting projects against the criteria does take some time and the TOC has recently modified the Due Diligence process to reduce duplication of work, streamline handoffs, and provide better transparency to the project and its adopters about its conformance and implementation of the criteria.
 
+##  Naming conventions for projects
+
+All CNCF projects are subject to the [Trademark Usage Policies](https://www.linuxfoundation.org/legal/trademark-usage) set by Linux Foundation. Specifically, new or incoming projects should avoid use existing trademarks in their proposed project names. In addition, if they are intending to use a popular prefix/suffix of an existing project (like "kube" or "k8s"), then they should consult the leadership group of the respective project to seek their approval and document the consensus reached. Existing projects are encouraged to document their naming guidelines to make this process smooth as well to avoid lengthy deliberation process for new project names.
```
linsun, line 28: AFAIK, k8s is a tademark of kubernetes so should be avoided.

I think we should ask new/incoming projects to follow existing project's naming guideline and not use the reserved prefix/suffix based on the guideline if the project has a naming guideline.  Consult the project leadership could be perceived as subjective as some project could be mostly owned by a single company.

linsun: s/existing trademarks/[existing rademarks](see https://www.linuxfoundation.org/legal/trademarks)

dims: @linsun `k8s` and `kube` is not trademarked. We have to cover the cases where the naming doc is not yet created by the leadership (Steering, TC or equivalent), so we have to cover that as well. I fully expect k8s steering folks to draft something soon, we just have to work with other projects to do theirs.

We should not be trying to deal with `single company` here. there are other mechanisms in our arsenal to deal with that i think.

dims:
xref: for those who are not yet aware, the trigger for this discussion/policy was https://github.com/cncf/toc/issues/1484

process/README.md
krook: FYI, on the last TOC call of December we did confirm that "K8s®" _is_ in the list of trademarks on https://www.linuxfoundation.org/legal/trademarks. The assumption was that it was not. "Kube" alone is not trademarked according to the list.

dims: @krook ah thanks for confirming!

BenTheElder: > Consult the project leadership could be perceived as subjective as some project could be mostly owned by a single company.

> We should not be trying to deal with single company here. there are other mechanisms in our arsenal to deal with that i think.

+1, IMHO single-company-owned projects is something that needs addressing regardless and shouldn't be in scope here.

(FWIW in this particular example, that's not the case, as we have [maximum representation rules](https://github.com/kubernetes/steering/blob/45c016ebe5c08cb3852240751382b02afc0aa4f3/elections.md?plain=1#L60) limiting to 2/7 of the committee ... I think those rules have worked well if only to at least limit _perception_ of excessive individual company influence.)

@dims is there anywhere tracking that topic currently?

>  I fully expect k8s steering folks to draft something soon, we just have to work with other projects to do theirs.

Yes, people are still filtering back from holiday vacations but I expect a draft soon. We meet tomorrow.

linsun: Cross link the PR from @BenTheElder https://github.com/kubernetes/community/pull/8238/files.

Okay with skipping single-company-owned projects.

linsun (approved):
lgtm

angellk (approved):
Thanks @dims - lgtm

PR #1523, opened 2025-01-15
fix: app delivery leadership

roberthstrand:


angellk (approved):
thanks @roberthstrand !

Issue #1532, opened 2025-02-10
Update project moving level guildance based on feedback from in-toto DD review


linsun:
A few things that come out of the in-toto review and worth clarify in the graduation template:

- as a specification project - there really is very little additional development that would need to happen.
- We do not hold projects back from moving levels on factors outside the project’s control that are probabilistic occurrences because many factors can change at a moments notice.
- for specification project, it is required to have at least one implementation and that reference implementation DOES NOT need to be part of the project undergoing due diligence.
- For graduated projects all critical and high issues from the security audit must be resolved.  For medium & low items - they should (not a requirement but strongly encourage) have a corresponding issue opened on the repo to track it to closure.

@TheFoxAtWork let me know if i missed anything, i'll open a PR for this shortly!

linsun:
Corresponding slack discussion in TOC private: https://cloud-native.slack.com/archives/G01MNU51LFM/p1736305335718719

linsun:
>     * We do not hold projects back from moving levels on factors outside the project’s control that are probabilistic occurrences because many factors can change at a moments notice.
> 

Not exactly sure where to add this, so i left this out.  Going to open the PR shortly!

>     * For graduated projects all critical and high issues from the security audit must be resolved.  For medium & low items - they should (not a requirement but strongly encourage) have a corresponding issue opened on the repo to track it to closure.
> 
Confirmed this is already in the dd-toc-guide.md

TheFoxAtWork:
First 💕 thank you!  For the not holding back projects bit, I would put that at the end of the paragraph of line 166. It would fit perfectly.

linsun:
> First 💕 thank you! For the not holding back projects bit, I would put that at the end of the paragraph of line 166. It would fit perfectly.

Awesome, PR is updated: https://github.com/cncf/toc/pull/1533/files PTAL! Thanks Emily!

angellk:
@TheFoxAtWork @linsun this is adding requirements for specification projects to the DD guide - where are those requirements initially captured and reviewed? From what I recall, the private-toc thread was initiated because the TOC does not have requirements specified. However, if that is inaccurate - it should be linked in the linked PR.

TheFoxAtWork:
[...]

PR #1533, opened 2025-02-10
Update dd guide for specification project

linsun:
Fixes https://github.com/cncf/toc/issues/1532


JustinCappos (approved):
Great change!   Thanks for the update here!

operations/dd-toc-guide.md
```diff
@@ -163,7 +163,7 @@ As an example, let's say a project is asserting their sub-projects have leadersh
...
+It is expected that the TOC's evaluation of a project's completion of the criteria may reveal a mismatch in understanding or an unexpected implementation. Documenting the TOC evaluation in the Due Diligence PR provides the project, TAGs, community, adopters, and TOC with a point of reference to understand if the criteria are meeting the outcomes required of a project for a certain maturity level, or if compensating mechanisms that supplement or augment the criteria are in place that work best for that specific project. We do not hold projects back from moving levels on factors outside the project’s control that are probabilistic occurrences because many factors can change at a moment's notice.
```
angellk, line 166: > We do not hold projects back from moving levels on factors outside the project’s control that are probabilistic occurrences because many factors can change at a moment's notice.

This language is a bit confusing here. I understand it was taken verbatim from a private TOC thread, however, the context does not quite make sense here.

operations/dd-toc-guide.md
```diff
@@ -179,6 +179,10 @@ In the course of conducting a project's due diligence, you may become aware of l
...
 
+#### Specification project
+
+If a project is a specification project such as the TUF, SPIFFE and in-toto projects, there really is very little additional development that would need to happen and it is reasonable for the specification project to have only a few maintainers. For a specification project, it is required to have at least one implementation and that reference implementation DOES NOT need to be part of the project undergoing due diligence.
```
angellk, line 184: > For a specification project, it is required to have at least one implementation and that reference implementation DOES NOT need to be part of the project undergoing due diligence.

These requirements should be linked here.

angellk (changes requested):
Need clarification on where the requirements of a specification project are capture - to link here. Also, removal of the first addition - does not make sense in this context.

operations/dd-toc-guide.md
linsun: Yes I was having a bit trouble to understand where this would fit, so I consulted @TheFoxAtWork and she suggested in here: https://github.com/cncf/toc/issues/1532#issuecomment-2649706362

operations/dd-toc-guide.md
linsun: What requirements?  i don't think we have any existing requirement doc for spec project.

angellk: @linsun that would need to be created before referencing the requirements in the DD guide

operations/dd-toc-guide.md
angellk: I understand .. it still is confusing here as it is out of context.

operations/dd-toc-guide.md
TheFoxAtWork: Okay digging through the TOC repo and some historical items i have access to from past specification projects, it appears we're missing core requirement for specification projects to have at least one reference implementation with sufficient adoption. It is derived from our principles: 
https://github.com/cncf/toc/blob/main/PRINCIPLES.md#important-principle-of-interoperability
> CNCF values this highly: that for any given ‘specification’, multiple implementations exist. Those implementations will use the project’s specification as the source of truth.
and 
https://github.com/cncf/toc/blob/main/PRINCIPLES.md#how-cncf-works-with-interface-definitions-like-cni-vs-standards-efforts-like-oci
> For the avoidance of doubt: Any written “specification” materials in CNCF shall have the same status as CNCF open source software and project technical docs. That is to say that specifications shall be updated and release versioned according to the conventions of a CNCF open source project.

So a couple of items. We the TOC know this requirement exists and yet we've not captured it correctly to date.  i think for now this language is fine but we need an issue to track the specification criteria and expectations so we can be more clear with projects as well as our selves.

since this is for the TOC DD guide, we should reference the principles here as "alluding" to the need for at least one reference implementation with sufficient adoption to assess maturity and viability of the specification.

TheFoxAtWork (commented):
clarified the reference implementation and called out a need for specific example.

JustinCappos:
>
> since this is for the TOC DD guide, we should reference the principles
> here as "alluding" to the need for at least one reference implementation
> with sufficient adoption to assess maturity and viability of the
> specification.
>

It may make to clarify this as the 'union of all implementations' having
sufficient adoption.    In many successful specs, different adopters will
implement their own copy for a variety of reasons.   So, each individual
implementation may have limited diversity of maintainers, adoption, etc.
but as a whole it can be broad and diverse.

TheFoxAtWork:
@JustinCappos  good call out, thank you for adding this. @linsun this will also need incorporated.

operations/dd-toc-guide.md
linsun: I've removed it for now

linsun:
> @JustinCappos good call out, thank you for adding this. @linsun this will also need incorporated.

Yep thanks to both, added now.

linsun:
@angellk do you want to take another look? thanks!

TheFoxAtWork (approved):

angellk (approved):
lgtm - thanks for the changes

PR #1535, opened 2025-02-14
fix: update template-incubation-application.md and template-graduation-application.md to fix the link for dd-toc-guide.md

avishnu:
The relative path link is not resolving when issue is created using this template.


angellk (approved):
thanks @avishnu !

PR #1574, opened 2025-03-26
Add "Initiatives" to README files and improve slack URL display

riaankleinhans:


riaankleinhans:
The automation did not react as expected, as there are other files change with the tags.yaml file. 
This would need to be fixed going forward.
This PR is good to merge, and would be followed by another PR to test the "forked branch" automation when only the tags.yaml are updated.

PR #1582, opened 2025-03-28
Update gitvote voters to use cncf-toc-voters

mrbobbytables:
Changes the voter group to `cncf-toc-voters`
As the shadows are included in the `cncf-toc` group it should no longer be used for voting.

The file was also "windows" format with the windows newline character at the end. I changed it to unix.


angellk (approved):
lgtm - ty!

PR #1583, opened 2025-03-30
Achive files and update readme

riaankleinhans:
Moving old tag files and folders in to an archive folder + update the TOC readme with the new TAGs


riaankleinhans:
Replaced by #1586

PR #1742, opened 2025-06-02
Fix tag devex slack link

dukex:


riaankleinhans:
@dukex thank you for this.
The update of the TAG READMEs are handled vir updating the [tags.yaml ](https://github.com/cncf/toc/blob/main/tags.yaml) file. You can follow the [README](https://github.com/cncf/toc/blob/main/generator/README.md) here.
However the error was created through a template error which I fixed here: https://github.com/cncf/toc/pull/1761

PR #1758, opened 2025-06-24
Update dd-toc-guide.md

linsun:
correct links


angellk (approved):
thanks @linsun ! lgtm

PR #1765, opened 2025-07-02
Add meeting agenda doc to README in TAG DevEX

danieloh30:


dims:
@danieloh30 there are extra files in the PR

danieloh30:
> @danieloh30 there are extra files in the PR

Good catch! Removed and updated .gitingore

riaankleinhans:
@danieloh30 as discussed, the README should be [updated](https://github.com/cncf/toc/tree/main/generator#readme) via the [tags.yaml](https://github.com/cncf/toc/blob/main/tags.yaml) 
There is no field for "agenda doc" in the tags.yaml

I would propose link the agenda doc in the meeting invitation in PCC.

PR #1773, opened 2025-07-15
Update EMERITUS according to last TOC election

pacoxu:
This is a follow up of https://github.com/cncf/toc/pull/1551/.

cc @mrbobbytables


mrbobbytables:
Thanks!

PR #1796, opened 2025-07-29
Add label action & configuration

mrbobbytables:
This PR defines the labels for the repo as well as enables a series of slash commands and other automation that can be used with triage.

unrelated, but it also removes the stalebot action that hasn't ever actually been used


kfaseela (approved):

.github/labels.yaml
```diff
@@ -0,0 +1,808 @@
...
+  color: ededed
+- name: vote/closed
+  description: An election that has been completed
+  color: 5c2908
+- name: vote/nomination
+  description: A nomination or call for nominations
+  color: E00EEF
+- name: vote/open
+  description: An election is open
+  color: 0e8a16
```
TheFoxAtWork, line 202: Is it possible that these labels may be confused with the TOC voting on projects for moving levels?  the description is clear, but the description isnt prominent via the label through the UI interface.

.github/labels.yaml
```diff
@@ -0,0 +1,808 @@
...
+      label: sub/{{ argv.0 }}
+
+
+##############################################################################
+# DD commands
+##############################################################################
+
+# triage for DD only includes items that are 'blockers'. Any project that has a 
+# complete app, submitted adopters, and has completed their security assessment
+# is ready to be evaluated and dd/status/ready-for-review
```
TheFoxAtWork, line 479: This needs matched with the TOC triage guidance: https://github.com/cncf/toc/blob/main/operations/dd-toc-guide.md#initial-triageevaluation-prior-to-assignment. the TOC Triage guidance is more comprehensive than these three items.

TheFoxAtWork: This will likely require a new PR to ensure they match/ the TOC Triage guidance is updated.

.github/workflows/labeler.yaml
```diff
@@ -0,0 +1,52 @@
...
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        
+      - name: Get changed files
+        if: ${{ github.event.pull_request }}
+        id: changed-files
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c
```
TheFoxAtWork, line 25: love that we're pinning to the commit SHA for this... How often do we plan to review this and update it?  Do we have an update policy for pinned dependencies?

TheFoxAtWork (approved):
Comments provided are for clarification and should not block the merging of this PR however there will be other issues that are spawned as a result of this so please be aware and ensure they are created.

.github/labels.yaml
mrbobbytables: Most of the items on that list I would consider fall under `dd/triage/incomplete-application` and would require additional comments in the issue. I could change it to not-enough-information? 

The others 2 are the other big ticket items: none or not enough adopters submitted, no security self assessment or completion of 3rd party security audit.

When those are 'done' its usually ready to picked up by a TOC member

.github/workflows/labeler.yaml
mrbobbytables: Ad hoc :p

the reason the other ones also reference main branch instead of a commit is in case we have to fix bugs, and we can do it without having to pen a 2nd PR here. Once we have things quite proven, we'll switch to commit or version.

.github/labels.yaml
mrbobbytables: I don't think so? they have to be manually applied, and someone would have to do that. We also use gitvote for those. These are for the other things such as leadership elections. 

If it'd help I can specify non moving levels related elections

.github/workflows/labeler.yaml
TheFoxAtWork: cool - just want to make sure we have a regular cadence on this so we dont inadvertently get caught up in a vulnerability because we're grabbing latest (even though our risk exposure is probably low)

.github/labels.yaml
TheFoxAtWork: Apologies my comment was less about what the label should be called, and more around 3 != 29 other items therefore we need to align these in our operation's triage guide. We provide links to how the TOC does things to maintainers and community members, it is IMPERATIVE that our labeling schema is reflecting in those docs so anyone can reason and understand the why.

If we're considering those other items as "incomplete" or "not-enough" that is fine, we just need to ensure its clear in the triage documentation we have and incorporates the right label to affix.  potentially, a project could have all three labels affixed so the key is to ensure during Triage, we complete looking at all aspects to label correctly.

.github/labels.yaml
TheFoxAtWork: So while i understand what they are for (the description says it here) when reviewing issues (particularly when not looking at a single issue, but an overview of several) that have the label on them it may not be apparent what the vote is about without viewing the direct issue in question. 

nomination is clear. but vote/open and vote/closed are not clear. it is just that a vote is occurring or has occurred.

we can use these generally but if we're talking about elections, we may want to be explicit about it.

.github/labels.yaml
mrbobbytables: Agreed, I absolutely think that all 3 will be on there, (the triage labels allow multiples vs the status ones which only allows one)

We can also add more labels to reflect triage state if thats a better option. Just..preferably not 29 of them XD

mrbobbytables: thought - would `needs-more-information` be better than `incomplete-application` ?

TheFoxAtWork: I think it depends on the tone of the message.

needs-more-information is a lot friendly than incomplete-application, so i'm in favor of the former.

mrbobbytables: updated, should be good to merge

mrbobbytables:
Since there have been several approvals and the outstanding change regarding needs-information has been commited, going to go ahead and merge this.

.github/workflows/labeler.yaml
evankanderson: You can use dependabot for this -- it can send PRs for these if you have a comment like `# @v46`, at least.

I can send a PR for dependabot configuration if desired.

.github/workflows/labeler.yaml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          separator: ","
+          
+      - name: Run labeler action for PR
+        if: ${{ github.event.pull_request }}
+        uses: cncf/automation/.github/actions/labeler-action@main
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          labels: ./.github/labels.yaml
```
evankanderson, line 35: This seems to be expected to be a URL: https://github.com/cncf/automation/blob/main/utilities/labeler/main.go#L101

```
 2025/08/06 21:36:44 failed to load labels.yaml: failed to fetch labels.yaml from URL: Get "./.github/labels.yaml": unsupported protocol scheme ""
```

Maybe this needs to be `file://./.github/labels.yaml`?

Would you like a PR?

PR #1810, opened 2025-08-07
Fix labeler workflow config

mrbobbytables:
Fixes the broken workflow by updating the label config path to use a url instead of filepath


evankanderson (approved):
I think you could also use `file://`, but this should work too.

mrbobbytables:
> I think you could also use `file://`, but this should work too.

yeah, I'll test that one later to verify in our automation repo. :+1: 
This one I KNOW will work and I'd rather get it up and going.

PR #1812, opened 2025-08-07
Update company affiliations, emails, and LFX IDs for TAG Security and Compliance

evankanderson:
Most important: Brandt Keller's name and GitHub id were spelled wrong.  Additionally, added company affiliations and emails (from git commit histories) in case those are being synced elsewhere.  (Otherwise, maybe we don't need them checked in?)


evankanderson:
Will reopen to get #1813 re-created.

PR #1818, opened 2025-08-12
Update README generator to fix charter overwrite and incorrect links

evankanderson:
Fixes three issues (because I got in there and got on a roll):

* The generator script was overwriting `charter.md` even if it had useful contents (examples: #1813, #1817)

* The `tags.yaml` links for TAG and Subproject initiatives were not working in all cases.  Given the uniformity in the repo, there's no point in making humans maintain this, so I auto-generated the links.

* The PR automation was not using the `-s` (signoff) flag on commits, so the generated commits were failing DCO (examples: #1813, #1817).  I'm still not 100% comfortable with this approach, as we're _copying_ the contents of the PR without providing original author attribution, but I'm picking this as a stopping point for now.

### DIFF
--- a/tags/tag-security-and-compliance/README.md
+++ b/tags/tag-security-and-compliance/README.md
@@ -1,5 +1,7 @@
 # TAG Security and Compliance
 
+<!-- THIS FILE IS AUTO-GENERATED FROM /tags.yaml -->
+
 ## Mission Statement
 Security hygiene, Policy-as-code, Compliance, Auditing, Threat Modeling, Secure Software Supply Chain
 
--- a/tags/tag-security-and-compliance/README.md
+++ b/tags/tag-security-and-compliance/README.md
@@ -12,9 +12,9 @@ Security hygiene, Policy-as-code, Compliance, Auditing, Threat Modeling, Secure 
 - Evan Anderson (**[@evankanderson](https://github.com/evankanderson)**), Custcodian (Term: 2025-07-01 - 2026-06-30)
 - Marina Moore (**[@mnm678](https://github.com/mnm678)**), Edera (Term: 2025-07-01 - 2027-06-30)
 ### Tech Leads
-- Brandt Keller (**[@brandtkeller](https://github.com/brandtkeller)**), Defense Unicorns (Term: 2025-07-02 - 2026-06-30)
+- Brand Keller (**[@brandkeller](https://github.com/brandkeller)**) (Term: 2025-07-02 - 2026-06-30)
 - Jennifer Power (**[@jpower432](https://github.com/jpower432)**), Red Hat (Term: 2025-07-02 - 2026-06-30)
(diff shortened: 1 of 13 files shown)


tags/tag-security-and-compliance/README.md
```diff
@@ -12,9 +12,9 @@ Security hygiene, Policy-as-code, Compliance, Auditing, Threat Modeling, Secure
 - Evan Anderson (**[@evankanderson](https://github.com/evankanderson)**), Custcodian (Term: 2025-07-01 - 2026-06-30)
 - Marina Moore (**[@mnm678](https://github.com/mnm678)**), Edera (Term: 2025-07-01 - 2027-06-30)
 ### Tech Leads
-- Brandt Keller (**[@brandtkeller](https://github.com/brandtkeller)**), Defense Unicorns (Term: 2025-07-02 - 2026-06-30)
+- Brand Keller (**[@brandkeller](https://github.com/brandkeller)**) (Term: 2025-07-02 - 2026-06-30)
```
mnm678, line 17: Is this intentionally being overwritten? The previous version is correct, but I'm not sure if this should be done elsewhere for the automation

eddie-knight: @brandtkeller for laughs 👀  ^

tags/tag-security-and-compliance/README.md
```diff
@@ -12,9 +12,9 @@ Security hygiene, Policy-as-code, Compliance, Auditing, Threat Modeling, Secure
...
 - Jennifer Power (**[@jpower432](https://github.com/jpower432)**), Red Hat (Term: 2025-07-02 - 2026-06-30)
-- John Kjell (**[@jkjell](https://github.com/jkjell)**), ControlPlane (Term: 2025-07-02 - 2027-06-30)
+- John Kjell (**[@jkjell](https://github.com/jkjell)**) (Term: 2025-07-02 - 2027-06-30)
```
eddie-knight, line 19: Same as above, is this a git history problem, or is something being generated that overwrites this?

tags/tag-security-and-compliance/README.md
evankanderson: See https://github.com/cncf/toc/pull/1813 (referenced in the examples).

Let me update the template to add a "THIS IS AUTOGENERATED" header, which I usually add as a best practice, but forgot to touch here.

tags/tag-security-and-compliance/README.md
evankanderson: See also #1813 / #1812

evankanderson:
I think this is probably @cncf/cncf-projects to approve, since it's mostly the generator script.

riaankleinhans (approved):

riaankleinhans:
@evankanderson thank you for improving this.

PR #1997, opened 2025-12-18
make copacetic GTR render in GitHub

joshgav:
The GTR doc for `copacetic` does not have the `.md` extension, causing it not to render in GitHub's interface. It will be easier for reviewers to read the review if it renders in GitHub. This PR adds that extension to do so. It also removes an extraneous `#` from the first line :wink:


brandtkeller (approved):
Scope of the change feels reasonable given the context. Maybe room for a rename to the expected format in this or future iterations.

angellk (approved):

PR #2036, opened 2026-02-04
Refactor tech review issue handling in workflow.

riaankleinhans:
- Removed redundant comment marker for existing tech review issues.
- Enhanced logging for issue processing, including issue labels.
- Implemented retry logic with backoff for checking existing tech review issues to prevent duplicates.
- Improved checks for identifying tech review issues based on title and body content.

### DIFF
--- a/.github/workflows/create-tech-review.yml
+++ b/.github/workflows/create-tech-review.yml
@@ -43,7 +43,6 @@ jobs:
 
             const COMMENT_MARKERS = {
               techReviewCreated: 'Created tech review issue:',
-              techReviewExists: 'tech review issue already exists',
               missingProjectName: 'Could not extract project name',
               missingProjectLink: 'Could not extract project link',
               createTechReviewFailed: 'Failed to create tech review issue'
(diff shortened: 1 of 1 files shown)


riaankleinhans:
@joshgav @kevin-wangzefeng this workflow automatically create a tech review issue.
The idea is that as soon as the `Tech-Review` label is added to the moving levels issue the workflow automatically create the tech review issue.
Due to an error in the logic it keep making duplicate tech review issues when ever you make a tech review issue manually.

Going forward you could just add the tech review label to the moving levels issue and the workflow would make the tech review issue.

Should you choose to make a manual tech review issue the workflow should not create a duplicate issue after this PR merged.

Is this workflow helpful? Or more in the way?

jeefy (approved):

joshgav:
Thank you @riaankleinhans for your work on this. The workflow is helpful but we need to keep iterating on it.

The requirement is that when a "moving levels" issue (and only a "moving levels" issue) is tagged with `review/tech` and `review/governance` the corresponding "tech review" and "gov review" issues are created with metadata prepopulated from the "moving levels" issue. When the "tech review" and "gov review" issues are themselves tagged with `review/tech` and `review/governance` it shouldn't trigger the workflow (which would create a duplicate).

Another requirement at some point is to add the issues to this board: <https://github.com/orgs/cncf/projects/88>, which is where we are tracking reviews work. In the meantime we can add them to the board manually.

Today we create tech/gov review issues manually cause the workflow hasn't worked correctly - when the "moving levels" issue is tagged `review/tech` the workflow simply posts a message to the same issue saying the workflow failed. The workflow actually only works when we tag the review issue itself :open_mouth: - it then creates a duplicate. Hopefully this PR fixes that! It's good that it won't create another issue if one has already been created as we plan to tag the review issues themselves with `review/tech`.

While we're discussing this here's something I'd like your thoughts on - is there a way we can prompt the person who opens the original "moving levels" issue to answer questions for the child "review" issues? For example, it would be ideal if we could move some of the questions from the "moving levels" issue itself into the reviews issues. However, it's important that the submitter be prompted to answer them. We don't have to deliver this requirement yet, but wanted to see if you have ideas on how we might approach it.

Thanks!

PR #2114, opened 2026-04-16
Require self-assessed technical and governance reviews for matriculation

joshgav:
This PR proposes requiring preliminary self-assessed technical and governance reviews for incoming and matriculating projects. It's based on consensus reached in the project-reviews subproject [on 2026-01-22](https://notes.cncf.io/77OV3RVUQd-dw3YkrJ7Leg#2026-01-22). We determined then that projects should complete their own technical reviews ("GTR") and that technical presentations to TAGs should be secondary and optional, with a process for such presentations yet to be determined. CNCF reviewers will then take a project's technical self-assessment, verify it and commit it to the cncf/toc repo, as they do for governance reviews today.

This PR also proposes a congruent process for governance reviews: a project will self-assess a governance review like the technical review prior to requesting matriculation. Projects already do exactly this today inline in [matriculation issues](https://github.com/cncf/toc/blob/main/.github/ISSUE_TEMPLATE/template-graduation-application.md), so this just asks them to save the assessment somewhere alongside their technical review rather than only including answers in the issue.

**Ultimately, the idea is that this will lead to technical, governance, and security reviews being saved and persisted by a project, e.g., in a `community` or `.project` repo, and referenced from matriculation issues**, rather than being inline in the matriculation issue itself. This will make review docs easier to maintain and iterate upon for projects and easier for users to discover.


JustinCappos:
+1

This is a welcome addition!  Hope to see it go through!

.github/ISSUE_TEMPLATE/template-graduation-application.md
```diff
@@ -40,16 +40,24 @@ _The project has been adopted by the following organizations in a testing and in
...
+  - This occurred on DD-MMM-YYYY and can be discovered at $LINK.
+
+<!-- (Project assertion goes here) --> 
 
 ### Required
 
 
-- [ ] **Engage with the domain specific TAG(s) to increase awareness through a presentation or completing a General Technical Review.**
-  - This was completed and occurred on DD-MMM-YYYY, and can be discovered at $LINK.
+- [ ] **Complete a [General Technical Review (GTR)](../toc_subprojects/project-reviews-subproject/general-technical-questions.md).**
```
brandtkeller: If I am reading the path correctly - should this instead be:


```suggestion
- [ ] **Complete a [General Technical Review (GTR)](../../toc_subprojects/project-reviews-subproject/general-technical-questions.md).**
```
(Also applies to the other references)

Alternatively - this is an issue template - does it have path-awareness for these files such that external viewers will be directed accordingly?

brandtkeller (changes requested):
Passing thought on issue template. Still reviewing updates.

angellk:
This needs to be discussed by the TOC - thank you for the recommendation from the Subproject @joshgav

angellk (approved):
Thank you! This was discussed by the TOC and this update is very appreciated @joshgav

brandtkeller (approved):

PR #2194, opened 2026-06-12
clarify security self-assessment location in DD guide and graduation …

kfaseela:
…template


JustinCappos (approved):
Super!  Thanks for this.  

If there is a similar place where the joint assessment is referenced, we should make a similar change.

mfahlandt (approved):
/lgtm
/approve

angellk (approved):
lgtm

PR #2201, opened 2026-06-17
Kubeflow Security Self Assessment

andreyvelich:
Ref: https://github.com/kubeflow/community/pull/996#issuecomment-4734302868
Adding initial Kubeflow security self-assessment

/cc @kfaseela @franciscojavierarceo @juliusvonkohout @chasecadet @thesuperzapper


projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,517 @@
+# Kubeflow Security Self-Assessment
+
+This document is a Kubeflow security self-assessment.
+
+Authors:
+
+- Kubeflow Community
```
JustinCappos, line 7: Can you list a few people so someone with questions knows who to reach out to / folks in the community 2 years from now recall who participated?

projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,517 @@
...
+
+[Kubeflow](https://www.kubeflow.org/) is the foundation of tools for AI Platforms on Kubernetes.
+
+AI platform teams can build on top of Kubeflow by using each project independently or deploying the
+entire AI reference platform to meet their specific needs. The Kubeflow AI reference platform is
+composable, modular, portable, and scalable, backed by an ecosystem of Kubernetes-native
+projects that cover every stage of the [AI lifecycle](https://www.kubeflow.org/docs/started/architecture/#kubeflow-projects-in-the-ai-lifecycle).
+
+Whether you’re an AI practitioner, a platform administrator, or a team of developers, Kubeflow
+offers modular, scalable, and extensible tools to support your AI use cases.
```
JustinCappos: This is more like marketing speak and isn't precise about what the security properties of the system are meant to be.  What security properties is it meant to provide?  What trust assumptions are there?  What happens when these are violated, etc.?

projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,517 @@
...
+
+![ai-lifecycle](images/ai-lifecycle-kubeflow.drawio.svg)
+
+The actors for each Kubeflow project are explained in the following sections:
+
+### Kubeflow Spark Operator
+
+![spark-operator](images/spark-operator.png)
+
+- Spark Operator controller: A controller that watches for events of SparkApplication CRDs and acts on the watch events. It includes a submission runner that runs Spark submit for submissions received from the controller, and a Spark pod monitor that watches for Spark pods and sends pod status updates to the controller.
```
JustinCappos: Is the submission runner validated / trusted?

projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,517 @@
...
+  hyperparameter values that the hyperparameter tuning process has proposed. Suggestion is
+  responsible to manage algorithm service.
+
+- Trail controller: controller that watches events of Trial CRDs which manage one iteration of
+  hyperparameter tuning process. A Trial corresponds to one worker job instance with a list of
+  parameter assignments. The list of parameter assignments corresponds to a Suggestion.
+
+- Katib webhooks: Validates and mutates CRD resources to ensure they conform to Katib standards
+  and best practices. Katib also manages admission webhook to mutate metrics collector sidecar
+  container into Trial workers.
```
JustinCappos, line 253: Who runs these?  Are they trusted?  What if a malicious party gets into one of these controllers, what could they impact?

projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,517 @@
...
+
+- Spark Application submission: Users create SparkApplication CRDs, specifying configuration for the
+  Spark Job.
+
+- Spark Operator controller automatically performs appropriate actions to orchestrate Spark jobs on
+  Kubernetes
+
+- Security and Access Control: Spark Operator leverages Kubernetes RBAC for Spark drivers and
+  executors. This allows administrators to define who can create, modify, or delete SparkApplications
+  and associated pods within the specific namespaces, enabling proper multi-tenant isolation.
```
JustinCappos, line 340: So is the assumption that multiple, untrusted parties will be using the same KubeFlow experiment controllers, suggestion controllers, etc.?

projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,517 @@
...
+  - Python packages: https://github.com/kubeflow/pipelines/blob/master/sdk/python/requirements.txt
+  - FOSSA report: https://app.fossa.com/reports/295e1511-cf5e-405c-abc3-eddac2ef03ae
+
+All Kubeflow container images can be found here: https://github.com/orgs/kubeflow/packages
+
+The SBOMs for any released Kubeflow container image can be accessed using the following command:
+
+```
+docker sbom ghcr.io/kubeflow/trainer/trainer-controller-manager:v2.0.0-rc.0
+```
```
JustinCappos: You should consider moving these parts lower in the document.  The SBOM part could go in SDP or compliance.  The rest of the links probably can be in the appendix.

projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,517 @@
...
+Kubeflow projects do not currently claim compliance with specific security standards
+(e.g., PCI-DSS, ISO, GDPR). The Kubeflow projects follow open source and Kubernetes security best
+practices and is following the OpenSSF Best Practices. Kubeflow projects achieve the OpenSSF badges:
+
+- Kubeflow Spark Operator: https://www.bestpractices.dev/en/projects/10524
+- Kubeflow Notebooks: https://www.bestpractices.dev/en/projects/9942
+- Kubeflow Katib: https://www.bestpractices.dev/projects/9941
+- Kubeflow Trainer: https://www.bestpractices.dev/projects/10435
+- Kubeflow Model Registry: https://www.bestpractices.dev/en/projects/9937
+- Kubeflow Pipelines: https://www.bestpractices.dev/en/projects/9938
```
JustinCappos, line 477: good job on these!

JustinCappos (changes requested):
Overall, I feel like the self assessment document needs more of a security focus.  Right now, it's too focused on what Kubeflow is meant to do, instead of how Kubeflow responds when things go wrong.  There are clearly some good steps the project has taken from a security standpoint, so I think this can likely be addressed by updating the writing.

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: Sure, added folks who contributed to the doc.
cc @droctothorpe @franciscojavierarceo @HumairAK @juliusvonkohout @nabuskey @tarilabs @thesuperzapper @vara-bonthu

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: Yes, the runner should be trusted. @roburishabh @vara-bonthu @yuchaoran2011 @ChenYi015 @nabuskey to confirm here.

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: This is managed by the Katib controllers in the system namespace. These are trusted components, and it is the responsibility of platform administrators to harden them and ensure they are protected from compromise.

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: Good point, I refactored the overview to make it more security focused. WDYT @JustinCappos?

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: Not really. We isolate users by namespaces, so with the appropriate ACLs users should not be able to manage other users resources.

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: SGTM, moved.

projects/kubeflow/security-assessment/self-assessment.md
nabuskey: Yes, the submitter is the operator itself.

projects/kubeflow/security-assessment/self-assessment.md
juliusvonkohout: Please take a look at https://github.com/kubeflow/community-distribution#architecture

projects/kubeflow/security-assessment/self-assessment.md
juliusvonkohout: Please check out https://github.com/kubeflow/community-distribution#architecture and PSS restricted/baseline + networkpolicies

JustinCappos: > it is the responsibility of platform administrators to harden them and ensure they are protected from compromise.

This part (and things like this) need to be clear up front.  What the operator is responsible for doing is really important to surface.  This should be clear in the self assessment docs and also in reasonable places on the kubeflow site / docs.

projects/kubeflow/security-assessment/self-assessment.md
JustinCappos: okay, can you try to split this section out so that the security compartments are clearer?  In other words, if someone manages to compromise component X and that lets them get into component Y, they are effectively not isolated.  Be clear about this please as well as what parts are meant to be isolated.

Note also that sometimes if a component is compromised it can do things like cause a DoS (by failing to schedule jobs, etc.) but cannot read or write sensitive data.  That's good to note too

projects/kubeflow/security-assessment/self-assessment.md
JustinCappos: okay, kindly make things like this get updated in the assessment doc.  Someone should be able to treat it as self contained and get an idea of the security of Kubeflow...

projects/kubeflow/security-assessment/self-assessment.md
RobuRishabh: @JustinCappos @nabuskey when you say trusted here means:
1. the controller image hasn't been tampered with
2. the CRD fields have been validated by the admission webhook before the controller processes them, and
3. additional restrictions (Pod Security Admission, NetworkPolicy, ResourceQuota) are enforced by the cluster
The operator itself does NOT sandbox or restrict what a SparkApplication can request that's the cluster admin's responsibility. 

So can we add some of this detail to the doc? Something like: 
```
The submission runner is not a separate service but the controller itself executing spark-submit as a subprocess. User-supplied fields from the SparkApplication CRD are passed as CLI arguments. The controller trusts that admission webhooks and cluster-level policies (PSA, NetworkPolicy) have validated these inputs before they reach the controller.
```

projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,545 @@
...
+The actors for each Kubeflow project are explained in the following sections:
+
+### Kubeflow Spark Operator
+
+![spark-operator](images/spark-operator.png)
+
+- Spark Operator controller: A controller that watches for events of SparkApplication CRDs and acts on the watch events. It includes a submission runner that runs Spark submit for submissions received from the controller, and a Spark pod monitor that watches for Spark pods and sends pod status updates to the controller.
+
+- Spark Mutating Webhook: a Mutating Admission Webhook that handles customizations for Spark driver
+  and executor Pods based on the annotations on the Pods added by the controller.
```
RobuRishabh: Quick thought: we actually have 4 webhook endpoints in Spark Operator, not just the one mutating webhook described here:
1. ValidatingWebhook for SparkApplication — validates naming, spec consistency, and optionally ResourceQuota ([source](https://github.com/kubeflow/spark-operator/blob/master/internal/webhook/sparkapplication_validator.go))
2. ValidatingWebhook for ScheduledSparkApplication — validates scheduled job specs
3. MutatingWebhook for SparkApplication — sets default values on the CRD
4. MutatingWebhook for Pods (SparkPodDefaulter) — the one described here, injects volumes, env vars, sidecars, security contexts into driver/executor pods

The pod mutator (#4) is especially security-relevant since it can inject `hostNetwork`, `hostPath` volumes, `privileged` security contexts, and arbitrary sidecars into pods based on what the SparkApplication CRD specifies.

Can we expand this list? I think for a security reviewer, knowing all four endpoints and especially what they do NOT validate (images, hostPath, privileged containers, we rely on cluster PSA for those) would be useful.

Also worth noting: if I am not wrong, all the webhooks have `failurePolicy=Fail`, so if the webhook server goes down, no SparkApplications or Spark pods can be created. Secure (fail-closed) but an availability concern. Worth a brief mention? wdyt ?

projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,545 @@
...
+
+- Spark Application submission: Users create SparkApplication CRDs, specifying configuration for the
+  Spark Job.
+
+- Spark Operator controller automatically performs appropriate actions to orchestrate Spark jobs on
+  Kubernetes
+
+- Security and Access Control: Spark Operator leverages Kubernetes RBAC for Spark drivers and
+  executors. This allows administrators to define who can create, modify, or delete SparkApplications
+  and associated pods within the specific namespaces, enabling proper multi-tenant isolation.
```
RobuRishabh, line 340: A couple of ideas to make it more useful for security reviewers:

1. Step-by-step submission flow: Could we describe the actual security pipeline? Something like:
- User creates SparkApplication CRD → ValidatingWebhook checks format/quota → MutatingWebhook sets defaults → Controller picks it up → Controller builds spark-submit arguments from CRD fields → spark-submit creates driver pod → SparkPodDefaulter mutating webhook injects configs into driver → driver creates executor pods → SparkPodDefaulter injects configs into executors
This would help reviewers see exactly where security checks happen and where they don't.

2. ProxyUser: Should we mention that Spark Operator supports proxyUser for Spark impersonation? It allows submitting jobs on behalf of another user. Security reviewers might want to know it exists and what guardrails are around it (currently: none beyond RBAC on who can set the field).

3. Driver → Executor trust escalation: The driver pod gets its own ServiceAccount and can create executor pods directly (native Spark behavior, not operator-controlled). Essentially, a user who can create a SparkApplication gets indirect pod-creation privileges through the driver. Should we call this out?

projects/kubeflow/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,545 @@
...
+
+![spark-operator](images/spark-operator.png)
+
+- Spark Operator controller: A controller that watches for events of SparkApplication CRDs and acts on the watch events. It includes a submission runner that runs Spark submit for submissions received from the controller, and a Spark pod monitor that watches for Spark pods and sends pod status updates to the controller.
+
+- Spark Mutating Webhook: a Mutating Admission Webhook that handles customizations for Spark driver
+  and executor Pods based on the annotations on the Pods added by the controller.
+
+Detailed information can be found here in the official
+[Kubeflow Spark Operator docs](https://www.kubeflow.org/docs/components/spark-operator/overview/#architecture).
```
RobuRishabh, line 219: Hey, I noticed we don't mention ScheduledSparkApplication anywhere. It's a cron-like feature that creates SparkApplications on a recurring schedule. From a security perspective, there are a couple of things worth noting:
- A scheduled job can continue running after a user's access is revoked (until someone explicitly deletes the ScheduledSparkApplication CR). This is a potential persistence concern in multi-tenant environments.
- It has its own validating webhook.
- It follows the same trust model as SparkApplication otherwise.

Should we add a brief mention here? Even just a line saying it exists and inherits the same security model?

Also, we have a newer CRD called SparkConnect (persistent Spark session endpoint) with its own validator and defaulter webhooks. It's still relatively new, but should we include it to be thorough from an attack surface perspective? Or is it too early?

projects/kubeflow/security-assessment/self-assessment.md
juliusvonkohout: Lets just say pss restricted as default also standalone @andreyvelich

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: Added this info @JustinCappos.

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: We didn't get a chance to work on these last year @RobuRishabh.
Could you to propose changes to this document which clarify those CRDs?

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: Updated this info @JustinCappos @juliusvonkohout.

projects/kubeflow/security-assessment/self-assessment.md
andreyvelich: Added those, PTAL

projects/kubeflow/security-assessment/self-assessment.md
RobuRishabh: Sure

RobuRishabh: Here it is : https://github.com/andreyvelich/toc/pull/1

andreyvelich:
@JustinCappos I believe, I addressed all of your comments.
Let us know if that looks good!
/assign @JustinCappos

JustinCappos (approved):
Thanks.  This is much improved!

I think this would be a moderate amount of work to do a joint assessment on.  However, the team has been really effective and responsive at responding to queries.  So, if there is interest and sustained effort from the project side, my largest concern at this point would be to get enough qualified reviewers together with time / ability to keep up.

kfaseela (approved):

andreyvelich:
> Thanks. This is much improved!

Thank you very much for your review @JustinCappos!
Happy to followup in the future reports to enhance it.

Issue #1815, opened 2025-08-12
[Security Review] kcp


embik:
Project Name: kcp

Github URL: https://github.com/kcp-dev/kcp

CNCF project stage and issue (NA if not applicable): sandbox (https://github.com/cncf/sandbox/issues/47) (we are looking for this joint security review to apply for incubation)

Security Provider: no

- [ ] Identify team
  - [ ] Project security lead
  - [ ] Lead security reviewer
  - [ ] 1 or more additional reviewer(s)
  - [ ] Every reviewer has read [security reviewer guidelines](/community/assessments/guide/joint-assessment.md) and stated declaration of conflict
  - [ ] Sign off by facilitator on reviewer conflicts
- [ ] Create slack channel (e.g. #sec-assess-projectname)
- [ ] Project lead provides draft document - see [outline](/community/assessments/guide/joint-assessment.md)
- [ ] "Naive question phase" Lead Security Reviewer asks clarifying questions
- [ ] Assign issue to security reviewers
- [ ] Initial review
- [ ] Presentation & discussion
- [ ] Share draft findings with project
- [ ] Assessment summary and doc checked into /assessments/projects/project-name (require at least 1 co-chair approval)
- [ ] CNCF TOC presentation (if requested by TOC)

JustinCappos:
Where is your self assessment document?  (You can see the repo for many examples.)

embik:
My understanding was that we can do either a self-assessment or a joint assessment with TAG security. If that is not the case I apologize for the confusion on my part!

JustinCappos:
The first part of a joint assessment is to do a self assessment.  [This book](https://tag-security.cncf.io/community/assessments/Open_and_Secure.pdf) can help to guide you.

embik:
Okay, understood. Apologies for opening the issue then. Should this be closed and re-opened once we have the self-assessment?

eddie-knight:
This issue is fine and good — You can link the self assessment PR to this issue as a reminder for folks that it's all related to the incubation application

mnm678:
Closing this for now. Please re-open when you have the self-assessment.

embik:
Hi @mnm678, we have had a security self-assessment for a while now, but I forgot to reopen this issue. My bad. Here it is: https://docs.kcp.io/kcp/v0.30/contributing/governance/security-self-assessment/

/reopen

embik:
Hm, no Prow here and I cannot reopen it myself. Please re-open @mnm678 :-)

[...]

PR #2209, opened 2026-06-26
add security self-assessment for project kcp

SimonTheLeg:
add security self-assessment for project kcp

moved per request from https://github.com/cncf/toc/issues/1815#issuecomment-4791938235

### DIFF
--- a/projects/kcp/security-assessment/self-assessment.md
+++ b/projects/kcp/security-assessment/self-assessment.md
@@ -1,0 +1,264 @@
+# kcp Security Self-Assessment
+
+This document is intended to aid the project's maintainers, contributors, and users understand the project's security status and help identify points of improvement.
+
+## Table of Contents
+
+* [Metadata](#metadata)
+  * [Security links](#security-links)
+* [Overview](#overview)
+  * [Background](#background)
+  * [Actors](#actors)
+  * [Actions](#actions)
+  * [Goals](#goals)
+  * [Non-goals](#non-goals)
+* [Self-assessment use](#self-assessment-use)
+* [Security functions and features](#security-functions-and-features)
+* [Project compliance](#project-compliance)
+* [Secure development practices](#secure-development-practices)
+* [Security issue resolution](#security-issue-resolution)
+  * [Incident Response](#incident-response)
+* [Appendix](#appendix)
+
+## Metadata
+
+|                   |                                                  |
+|-------------------|--------------------------------------------------|
+| Assessment Stage  | Complete                                         |
+| Software          | [kcp](https://kcp.io)                            |
+| Security Provider | No                                               |
+| Languages         | Go                                               |
+| SBOM              | kcp does not currently generate SBOMs on release |
+| Security links    | See below                                        |
+
+### Security links
+
+| Doc                       | url                                                          |
+|---------------------------|--------------------------------------------------------------|
+| Reporting security issues | [here](https://github.com/kcp-dev/kcp/security)              |
+| Security process          | [here](https://github.com/kcp-dev/kcp/blob/main/SECURITY.md) |
+
+## Overview
+
+kcp is an open-source, Kubernetes-like control plane designed for multi-cluster and multi-tenant environments. It provides a control plane that is not tied to a specific set of hardware, allowing users and services to manage APIs across a group of "logical clusters". kcp is specifically NOT a control plane for managing container workloads but higher level abstractions, which in turn might be run on separate Kubernetes clusters.
+
+### Background
+
+The kcp project began as an experiment to add the concept of "logical clusters" to the Kubernetes API server, effectively decoupling the powerful control plane from its traditional role in container orchestration. The project is built upon upstream Kubernetes code but is not a fork, with a strong commitment to maintaining 100% compatibility with the Kubernetes API machinery and ecosystem tooling. This allows kcp to provide a generic, horizontally scalable control plane for use cases beyond containers.
+
+### Actors
+
+The kcp architecture is composed of several key primitives that work together to provide a scalable, multi-tenant control plane.
+
+#### kcp Server Components
+
+* **Workspaces**: The primary user-facing unit of tenancy and isolation. From a user's perspective, a Workspace is a fully-isolated, Kubernetes-like cluster with its own unique API endpoint, CustomResourceDefinitions (CRDs), and RBAC policies.
+
+* **Logical Clusters**: The underlying implementation construct for a Workspace. A logical cluster is a logical partition within the kcp data store (etcd), ensuring that objects from different workspaces are stored in disjoint key prefixes, which is the primary mechanism for enforcing isolation. The goal is to make creating a logical cluster as cheap and fast as creating a Kubernetes namespace.
+
+* **Virtual Workspaces**: Endpoints that provide a Kubernetes-like API interface, but are not backed by a logical cluster for storage. They provide a computed "view" of certain resources across logical clusters. The exact semantics depend on the virtual workspace implementation, different virtual workspace endpoints provide different views according to their role. Access to virtual workspace endpoint is guarded by RBAC.
+
+* **Shards**: A running instance of the kcp server process. Each shard hosts a set of logical clusters, and a full kcp installation can be composed of many shards to achieve horizontal scalability.
+
+#### kcp kubectl Plugin (CLI)
+
+kcp provides a set of plugins for `kubectl`, the Kubernetes command line client. These plugins interact with kcp through its Kubernetes-like APIs and offer command line tooling for interacting with the unique features of kcp (see above for some of them). All of the interactions are executed via Kubernetes client libraries and use the client credentials passed to `kubectl` (e.g. via the `KUBECONFIG` environment variable).
+
+##### kcp Components
+
+* **kcp**: The `kcp` binary provides the means to run a kcp shard (see above). It can either be launched completely standalone and then embeds a "mini" front-proxy, virtual-workspaces and a cache-server to run a fully functional kcp instance, or be run just to serve API endpoints and run controller loops.
+
+* **cache-server**: The cache server provides a shared Kubernetes-like API layer that all shards connect to and cache objects relevant to other shards (e.g. `APIExports`) so that shards do not need to directly interact with each other.
+
+* **kcp-front-proxy**: An intelligent API gateway that sits in front of all shards. It is aware of all logical clusters and the shards they reside on, and it routes incoming user requests to the correct shard.
+
+#### API Management
+
+kcp provides resources dedicated to managing available APIs in a Workspace.
+
+* **APIExport**: Allows a service provider in one workspace to publish an API for consumption by other workspaces.
+
+* **APIBinding**: Allows a service consumer in one workspace to bind to an APIExport from another workspace, making the published API available in the local workspace.
+
+ The ability to bind APIs across workspaces (a security boundary) is guarded by RBAC checks.
+
+### Actions
+
+* **Workspace Management**: Users can create, list, and navigate between workspaces, each providing an isolated API endpoint.
+
+* **API Sharing**: Service providers can securely offer APIs across workspace boundaries using a declarative APIExport and APIBinding model.
+
+* **API Request Routing**: The front-proxy routes all standard client traffic to the appropriate shard based on the workspace path in the request URL.
+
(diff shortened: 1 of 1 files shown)


projects/kcp/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,264 @@
...
+
+The kcp project began as an experiment to add the concept of "logical clusters" to the Kubernetes API server, effectively decoupling the powerful control plane from its traditional role in container orchestration. The project is built upon upstream Kubernetes code but is not a fork, with a strong commitment to maintaining 100% compatibility with the Kubernetes API machinery and ecosystem tooling. This allows kcp to provide a generic, horizontally scalable control plane for use cases beyond containers.
+
+### Actors
+
+The kcp architecture is composed of several key primitives that work together to provide a scalable, multi-tenant control plane.
+
+#### kcp Server Components
+
+* **Workspaces**: The primary user-facing unit of tenancy and isolation. From a user's perspective, a Workspace is a fully-isolated, Kubernetes-like cluster with its own unique API endpoint, CustomResourceDefinitions (CRDs), and RBAC policies.
```
JustinCappos: Can you explain more about how this is isolated from other actors?  How does one interface with this in a trustworthy manner?

projects/kcp/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,264 @@
...
+
+* **Logical Clusters**: The underlying implementation construct for a Workspace. A logical cluster is a logical partition within the kcp data store (etcd), ensuring that objects from different workspaces are stored in disjoint key prefixes, which is the primary mechanism for enforcing isolation. The goal is to make creating a logical cluster as cheap and fast as creating a Kubernetes namespace.
```
JustinCappos, line 57: Do you have other performance / side channel concerns as a result?  (This likely goes in non-goals for your system as well.)

projects/kcp/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,264 @@
...
+
+* **Virtual Workspaces**: Endpoints that provide a Kubernetes-like API interface, but are not backed by a logical cluster for storage. They provide a computed "view" of certain resources across logical clusters. The exact semantics depend on the virtual workspace implementation, different virtual workspace endpoints provide different views according to their role. Access to virtual workspace endpoint is guarded by RBAC.
+
+* **Shards**: A running instance of the kcp server process. Each shard hosts a set of logical clusters, and a full kcp installation can be composed of many shards to achieve horizontal scalability.
```
JustinCappos: How are these coordinated?  Are these trusted by other shards?  What happens if a shard is compromised / buggy?

projects/kcp/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,264 @@
...
+
+#### kcp kubectl Plugin (CLI)
+
+kcp provides a set of plugins for `kubectl`, the Kubernetes command line client. These plugins interact with kcp through its Kubernetes-like APIs and offer command line tooling for interacting with the unique features of kcp (see above for some of them). All of the interactions are executed via Kubernetes client libraries and use the client credentials passed to `kubectl` (e.g. via the `KUBECONFIG` environment variable).
+
+##### kcp Components
+
+* **kcp**: The `kcp` binary provides the means to run a kcp shard (see above). It can either be launched completely standalone and then embeds a "mini" front-proxy, virtual-workspaces and a cache-server to run a fully functional kcp instance, or be run just to serve API endpoints and run controller loops.
```
JustinCappos, line 69: Is this a trusted component?  How does this auth?

projects/kcp/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,264 @@
...
+
+* **cache-server**: The cache server provides a shared Kubernetes-like API layer that all shards connect to and cache objects relevant to other shards (e.g. `APIExports`) so that shards do not need to directly interact with each other.
+
+* **kcp-front-proxy**: An intelligent API gateway that sits in front of all shards. It is aware of all logical clusters and the shards they reside on, and it routes incoming user requests to the correct shard.
+
+#### API Management
+
+kcp provides resources dedicated to managing available APIs in a Workspace.
+
+* **APIExport**: Allows a service provider in one workspace to publish an API for consumption by other workspaces.
```
JustinCappos: What if the same API is published for consumption by different parties?  Is this namespaced somehow?  If so, how is the namespacing securely defined / allocated?  If not, what happens if there is a conflict?

projects/kcp/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Goals
+
+* Provide a scalable, multi-tenant control plane that can serve hundreds or thousands of isolated tenants.
+* Enable service providers to offer Kubernetes-like APIs that can be consumed by consumers across multiple isolated tenants.
+* Maintain a high degree of compatibility with the standard Kubernetes API and toolchain (e.g., kubectl, client-go, multicluster-runtime).
+
+#### Security Goals
+
+* **Strict Workspace Isolation:** Ensure that tenants in one Workspace cannot see, access, or affect resources in another Workspace unless explicitly authorized. This is the core security boundary of the system.
```
JustinCappos, line 103: I think performance interference + side channels are likely out of scope

projects/kcp/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,264 @@
...
+
+#### Vulnerability Response Process
+
+Maintainers provide a [Security Response Team](https://github.com/kcp-dev/kcp/blob/main/MAINTAINERS.md#security-response-team) that organizes the entire response, including internal communication and external disclosure.
+
+#### Reporting Security Vulnerabilities
+
+Security vulnerability reports are handled via GitHub's security issue reporting feature, available [here](https://github.com/kcp-dev/kcp/security). The Security Response Team triage and respond to security issues reported privately through the tool.
+
+Please see the complete [security release process](https://github.com/kcp-dev/kcp/blob/main/SECURITY.md) for further details.
```
JustinCappos, line 233: Looks good!

JustinCappos (commented):
This looks pretty good.  Can you address the comments I have and integrate those into the text (not just reply to my comments)?  

I think this will be ready to merge soon.

projects/kcp/security-assessment/self-assessment.md
SimonTheLeg: I am a bit stuck, English is not my native language: What is meant by "side channel concerns" in this context?

projects/kcp/security-assessment/self-assessment.md
SimonTheLeg: The single binary is mainly a convenience option for quickstarts and local development. It runs all the components in a single binary and uses loopback and a fixed token for authn and authz. But it is not meant to be run in production.

Under these circumstances, should we remove the single binary part? In production the mentioned components would be run individually.

SimonTheLeg:
I have added some of the requested info and bundled it in a separate commit for easier review: [`4b0a794` (this PR)](https://github.com/cncf/toc/pull/2209/changes/4b0a794efa86c67d1cc2f0ab378f091778dc4867) or added some follow up questions into the discussion. ptal

projects/kcp/security-assessment/self-assessment.md
JustinCappos: Sorry for being unclear.  There is a type of attack where two things (processes, containers, etc.) running on the same hardware can interfere with each others performance or read information about what computation another process is doing because they share the same hardware.  I was just wondering if this vector is a concern here.  I assume you consider this to be out of scope.

evankanderson:
@realshuting and @JustinCappos - I think the ball is in your court.

Santoshkumarpuppala (commented):
Thanks for putting this together — the structure is solid and the Actors section is unusually clear about what kcp is and isn't, which makes the boundaries much easier to reason about than most self-assessments.

I read it with one narrow lens, and it's the one I'd offer as the general suggestion: for each isolation boundary the document claims, does it name the single server-side point that enforces it, and what would a negative test look like — a request that should be refused, and is. The goals are described well throughout; the enforcement points and the negative tests are the thin parts, and they're where a reader's confidence actually comes from.

One observation first, because I think it's the most useful thing I can hand back. The two advisories listed under Known Issues land exactly on the two boundaries this document describes least concretely:

GHSA-w2rr-38wv-8rrp (unauthorized create/delete in arbitrary workspaces through the APIExport Virtual Workspace) maps onto the Virtual Workspace Isolation section, which is written as a requirement ("it is therefore critical that virtual workspace implementations properly isolate requests... and use proper impersonation and/or enhanced authN/authZ checks") rather than as a description of what enforces it.

GHSA-c7xh-gjv4-4jgv (impersonation allows access to global administrative groups) maps onto impersonation, which appears once in the document — as a suggested mitigation inside that same section — and is never listed as a security-relevant feature with a guard of its own.

That's the project's own incident history pointing at which boundaries need naming. I'd suggest linking Known Issues back to the relevant boundary sections rather than leaving them as a standalone appendix list, so a reader can see "this boundary failed once, here is what enforces it now."

Specific spots, in rough priority order:

1. Workspace Isolation — the assessment says logical clusters store objects "in disjoint key prefixes, which is the primary mechanism for enforcing isolation." Disjoint storage prefixes are a storage property; they prevent collision and accidental cross-reads, but they aren't what refuses a request. Since the workspace is derived from the URL path, the reader's real question is: when a user with valid credentials changes the path to a workspace they have no access to, which component denies it — the front-proxy, the shard's authorizer, or per-workspace RBAC — and at what point in the request chain? Naming that one component (and stating that the path is untrusted input) would do a lot for the "core security boundary."

2. Virtual Workspace Isolation — this is the section I'd most want expanded, both because the document itself flags it as critical and because of GHSA-w2rr-38wv-8rrp. It notes these components "often have to run with administrative permissions." Two things would help: whether isolation is enforced centrally or is each implementation's own responsibility (if the latter, that's N enforcement points rather than one, and worth saying plainly), and what a new virtual workspace implementation is required to do before it ships. A checklist item is more durable than an adjective.

3. Impersonation — given the prior advisory, I'd list it as a security-relevant feature in its own right: what restricts which users/groups can be impersonated, and specifically what prevents impersonating global administrative groups.

4. Administrative Wildcard Requests — this appears under Actions as an endpoint that lists resources across all logical clusters, "bypassing standard workspace isolation for operational purposes." An explicit, documented isolation bypass belongs in Security Functions with its guard named: who is authorized, what refuses everyone else, and whether use is audited.

5. Permission Claims and Maximal Permission Policy — both are described as consent/upper-bound mechanisms, which is the right design. What's missing is what evaluates them at request time. For PermissionClaims: what refuses a provider access that exceeds what the consumer accepted? For maximalPermissionPolicy: where is the intersection with the consumer's own RBAC computed? These read as declarations today; naming the evaluator makes them controls.

6. APIExport identity — the identity hash is described as part of the etcd storage path, "ensuring complete data segregation." Same distinction as (1): that segregates data at rest. Is the identity also verified on the request path, so a request referencing another provider's identity is refused rather than simply missing its data?

7. Cross-workspace binding — "guarded by RBAC checks" is the right answer; naming the specific verb and resource (which side the check runs on, provider or consumer) would make it verifiable by a reader.

8. Shard trust — each shard gets a client certificate at startup and shards self-register with the root shard via bootstrap credentials. The document addresses malfunctioning shards ("functionality of functioning shards is not affected"), which is an availability property. The authorization question is different: is a shard's identity scoped to the logical clusters it hosts, or can any authenticated shard assert authority over any logical cluster? Worth stating explicitly whether shards are inside the trust boundary by design — that's a legitimate design decision, it just needs to be visible.

Two smaller things:

The Integration Tests section describes CI coverage generally. Given the two known advisories, it would be worth saying whether security regression tests are pinned to them specifically — that's the "negative test" question in its most concrete form, and it's the single strongest signal that a fixed boundary stays fixed.

There's no explicit attacker model. The Security Goals are clear about what should hold, but not about what the adversary is assumed to be able to do — an authenticated tenant in a neighbouring workspace, a malicious API provider, a compromised shard. Even a short paragraph would make the rest of the document easier to evaluate, because most of the questions above resolve once the assumed adversary is stated.

None of this is a finding — it's all "the document asserts a boundary and I can't tell from the text what enforces it." Happy to be pushed back on any of it, especially where the answer is obvious to people who work in the codebase daily and just didn't need writing down.

evankanderson:
@JustinCappos and @realshuting -- are you comfortable approving this, or are there more changes to this self assessment that need to be made?

JustinCappos:
I have a scattered set of unresolved comments which should have changes applied in the text.  After this, I would approve.

PR #2251, opened 2026-07-23
docs(infralifecycle): suggestion for use cases section

NotAwar:
This commit is an initial suggestion for the use cases section. 

I already see some improvements which I´ll note for the future in this suggestion, as I have thought about it while writing. 

Some ideas for further writing:
* The use cases I have written about are quite generic and I actually think we should consider either adding "examples" to this section to give specific implementation advice or consider expanding the use-cases with implementation examples. For instance show which tools are part of an infrastructure team stack, and which tools are expected to be used by a platform team to manage the infrastructure lifecycle of the resources they create. 
* I would consider cutting the summary and broaden the landscape title 
* I would consider sectioning landscape to differentiate between popular CNCF tools (what the CNCF offers in the landscape), and what is generically open source, as well as what is licensed. There is a ton of tools I have not mentioned which are open source, closed source and licensed. For instance, tools like Spacelift which work to support in managing infrastructure lifecycle are probably worth mentioning even though it also might feel out of scope in the current draft.

### DIFF
--- a/tags/tag-infrastructure/initiatives/infrastructure-lifecycle/04_Landscape.md
+++ b/tags/tag-infrastructure/initiatives/infrastructure-lifecycle/04_Landscape.md
@@ -1,13 +1,118 @@
 # Use Cases & Landscape
 
+The following section exemplifies the previously described patterns. Each use case describes the main pattern, relevant projects, and the trade-offs that should be considered before adoption. The section also provides an overview of the CNCF landscape for infrastructure lifecycle management.
+
 ## Case 1: Declarative Tooling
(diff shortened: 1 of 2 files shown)


mesutoezdil (approved):
LGTM, merging as a starting point!

Thanks for this! I like how you grounded it in actual projects instead of staying abstract.
Going to merge this and keep building on it from here, some of the things you already flagged, more concrete examples we'll fold in later.
Appreciate you jumping in on this.

GenPage (approved):

PR #2268, opened 2026-08-06
docs(higress): add security self-assessment for incubation

EndlessSeeker:
## Summary

Adds the Higress project-maintained Security Self-Assessment at the canonical
path used for CNCF incubation due diligence.

- Original source: https://github.com/higress-group/community/blob/main/docs/cncf/security-self-assessment.md
- Earlier boundary clarification: https://github.com/higress-group/community/pull/2
- Latest review follow-up: https://github.com/higress-group/community/pull/7
- Security reporting policy update: https://github.com/higress-group/higress/pull/4624
- Follows the CNCF TAG Security self-assessment outline
- OpenSSF Best Practices status: Passing

## Review follow-up

The current revision:

- explains the architecture in less specialized language;
- clarifies that AI prompts are request-body data when Higress is used as an AI gateway;
- records incomplete project-wide SBOM, signature, and provenance-attestation coverage;
- identifies the enforcement points for critical trust boundaries;
- distinguishes `allowedRoutes.namespaces` from `ReferenceGrant`;
- clarifies plugin scoping and last-known-good behavior; and
- makes GitHub Private Security Advisories the required, vendor-neutral, and
  authoritative intake, with Alibaba ASRC available as an optional additional
  channel. ASRC access is not required for Security Response Team membership.

## Scope

This is a project self-assessment, not an independent audit or joint
assessment. The content is synchronized from the project-maintained source and
updated in response to review feedback on this PR. The revision changes
documentation and reporting governance only; it does not alter runtime
behavior.

## Checklist

- [x] DCO signed
- [x] Added as `projects/higress/security-assessment/self-assessment.md`
- [x] Source and public evidence links included
- [x] Review feedback incorporated

## Agent-assisted contribution

- Codex assisted with implementation analysis and documentation wording.
- Gate status: verified maintainer/administrator exception under the Higress
  agent-assisted contribution policy.
- PR: https://github.com/cncf/toc/pull/2268
- Authenticated GitHub login: `EndlessSeeker`.
- Canonical `higress-group/higress` role: `admin`.
- Actual PR author: `EndlessSeeker`.
- `GH_TOKEN` and `GITHUB_TOKEN` were unset for every identity, permission, and
  PR-author verification command.
- Bypass rationale: focused documentation clarification on an in-flight CNCF
  security-assessment review; no runtime behavior changes.
- The CNCF TOC PR template has no Higress verified-exception selector, so the
  status is recorded here.


Santoshkumarpuppala (commented):
This is the most security-mature of the self-assessments I've read in this batch, and I want to be specific about why: it ships an actual threat model with residual-risk columns, an actors/assets/trust-boundary table, and a Known Gaps section that volunteers the uncomfortable items (the commented-out linter, the broad controller ClusterRole, absent SBOM/signature/provenance, and the vendor-neutrality question around dual ASRC reporting). Flagging your own vendor-neutrality concern before a reviewer raises it is exactly the right instinct. Most of what follows is refinement, not a request to add missing sections.

I read it with one narrow lens: for each boundary the document claims, does it name the single server-side point that enforces it, and what would a negative test look like — a request that should be refused, and is. The threat model is strong on scenario and residual risk; where it's thinner is the line between what Higress itself refuses and what it delegates to Kubernetes RBAC or the operator's configuration. Several High rows resolve to "operator responsibility," which is often the honest answer for a gateway, but a reader can't always tell which half of the boundary is Higress's.

The row I'd start with is HG-TM-06 (cross-tenant route/policy attachment), since it's the one most readers will scrutinize for a shared gateway. The mitigation lists "Gateway API attachment controls" generically. Gateway API has named mechanisms for precisely this: a Gateway's allowedRoutes.namespaces selector decides which namespaces may attach Routes, and ReferenceGrant governs cross-namespace references to Secrets and backends. Those are the concrete points that refuse a Route in one namespace from binding a Gateway in another. Naming them — and saying whether Higress relies on the upstream Gateway API implementation or adds its own enforcement — would turn that row from "controls exist" into a testable statement: a Route from an unlisted namespace attaching to a protected Gateway, refused.

A few more, in rough priority:

1. Enforced-by-Higress vs delegated. It would help to mark, per critical boundary, whether enforcement is in Higress code, in Envoy, or in Kubernetes RBAC that the operator must configure. HG-TM-01, HG-TM-02, and HG-TM-06 all lean on "Kubernetes RBAC" as the mitigation — which is correct, but it means the boundary is only as good as the operator's RBAC, and that's worth stating as the explicit dependency rather than as a Higress control. The distinction matters most where the controller's own ClusterRole is broad (which Known Gaps already admits): a broad controller role widens what a compromised controller can do regardless of the operator's tenant RBAC.

2. The commented-out golangci-lint. Known Gaps notes it's disabled "because of existing findings." Worth calling this out slightly more sharply, because a disabled linter is a control that reads as present — it's in the CI config — while enforcing nothing. Either a tracked path to re-enabling it (even lint-new-code-only, so the baseline debt doesn't block it) or an explicit "not currently enforced" note would keep the secure-development section from overstating coverage. CodeQL plus the go vet gate are doing real work; the lint line is the one that currently claims more than it delivers.

3. Wasm plugin grants. The trust-boundary table says "plugins can still observe or modify traffic granted to them," which is the right framing. The follow-up a reader wants: what does the granting? Is there a capability/scoping model for what traffic or configuration a plugin may touch, or does a loaded plugin see all traffic on the routes it's attached to? HG-TM-03 treats plugin installation as the control point (explicit install), but once installed, the scope of a plugin's access is the residual question — and native filters, as you note, share the process outright.

4. Last-known-good and policy removal. HG-TM-04 frames "retain last accepted configuration" as an availability mitigation, which it is. There's a security corner worth a sentence: if the control plane is unavailable, does the removal or tightening of a policy (say, revoking an auth filter or narrowing a route) also fail to propagate, leaving a stale permissive config serving? Availability-preserving last-known-good and timely policy revocation can pull in opposite directions, and stating which wins under partition would close the question.

One process note, not a document issue: the dual-reporting requirement (GitHub Private Security Advisories plus Alibaba ASRC) that you've already flagged is the right thing to have surfaced. From a reporter's side, the coordinated-disclosure concern is simply that a reporter must file into a vendor-operated channel to reach the project at all; a neutral primary intake (GitHub PSA as the system of record, with ASRC correlated behind it) would resolve most of it without losing the vendor's response capacity. Sharing that only because it's the disclosure-process angle, and you've clearly already thought about it.

None of this is a finding — it's all "the document asserts a boundary and I'd like the text to name what enforces it, or the dependency it rests on." The threat model is a good base to be doing this against. Happy to be pushed back on any of it, particularly the Gateway API enforcement points, where your implementation details will be more current than my read of the spec.

EndlessSeeker:
@Santoshkumarpuppala Thanks for the careful review. We have updated the assessment to name the enforcement point for the critical boundaries, distinguish `allowedRoutes` from `ReferenceGrant`, state the current golangci-lint status, clarify plugin scope, and document the stale-policy tradeoff in last-known-good behavior; the project-maintained source update is higress-group/community#2. We are handling the dual-reporting policy separately with project maintainers, since it needs a governance decision rather than a wording-only change.

Santoshkumarpuppala:
Thanks — I've read the revision here and the source change that merged upstream. The HG-TM-06 rewrite is what I was asking for: it names allowedRoutes.namespaces and ReferenceGrant as two separate checks rather than one general control, and then says what each one refuses — a Route from an unselected namespace doesn't attach, and a cross-namespace reference without a matching grant is rejected. That last sentence is the part a reader can actually act on, because it's falsifiable.

The change I didn't ask for is the better one. HG-TM-03 now introduces matchRules and in the same paragraph says they scope intended execution but are not a security sandbox against a malicious installed plugin. Adding a mechanism to a threat model and disclaiming it in the same breath is the harder move, and it's what stops a reader assuming that scoping is isolation.

On dual reporting — treating it as a governance decision rather than a wording fix is the right call. A reporter needs to know which intake is authoritative before they send anything, and that isn't something a self-assessment can settle on its own.

EndlessSeeker:
Hi @angellk — the review feedback has been addressed, the upstream source update is merged, and DCO is passing. The PR is conflict-free and appears ready for liaison review. Could you take a look or point us to the right approver? Thanks!

projects/higress/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,283 @@
...
+  identity metadata, certificates, or AI/model APIs when configured.
+- **Project maintainer/release manager:** reviews changes, responds to reports,
+  and publishes releases.
+
+### Background
+
+Higress combines the Envoy data plane and Istio-derived control plane with
+Ingress/Gateway API translation and an extensible plugin ecosystem. The
+control plane watches configuration and discovery sources and generates xDS;
+the data plane accepts untrusted network traffic and applies that configuration.
```
JustinCappos: Cam you write this in a less jargon-y way?  Maybe twice the length as what oyu have here.

projects/higress/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,283 @@
...
+| HG-TM-06 | High | A tenant accidentally or deliberately attaches a route or policy to another tenant's gateway and exposes or disrupts traffic. | Kubernetes RBAC and admission policy control who may create or change resources. Higress's Gateway API translation enforces a listener's `allowedRoutes.namespaces` for Route-to-Gateway attachment and separately checks `ReferenceGrant` for cross-namespace backend and Secret references. A Route from a namespace not selected by `allowedRoutes` does not attach, and a cross-namespace reference without a matching `ReferenceGrant` is rejected. Higress does not make a hostile multi-tenant cluster safe when administrators grant overlapping write privileges. |
+| HG-TM-07 | High | Requests, AI prompts, credentials, or responses are disclosed to an external registry, plugin, observability backend, identity service, or model provider. | Integrations require operator configuration and can be restricted through credentials and network policy. Operators remain responsible for egress allowlists, provider contracts, log redaction, residency, and data-retention controls. |
```
JustinCappos: Where are the AI prompts used in Higress here?  I don't see this as a component

projects/higress/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,283 @@
...
+| HG-TM-08 | Medium | Logs, metrics, traces, admin/debug endpoints, or configuration dumps expose credentials or sensitive request metadata. | Admin interfaces are not intended as public endpoints and telemetry is configurable. Access control, redaction, retention, and exposure are deployment responsibilities; unsafe logging or endpoint exposure remains possible. |
+| HG-TM-09 | High | Certificate expiration, issuer compromise, or unauthorized Secret replacement causes outage or impersonation. | SDS and automatic HTTPS support dynamic certificate updates and renewal. Operators must secure issuers, monitor expiry/renewal, test emergency rotation, and control Secret writers. |
+| HG-TM-10 | High | A source-control, dependency, GitHub Actions, registry, or maintainer-account compromise produces malicious release artifacts. | Public review, CI tests, license checks, CodeQL, pinned dependency/submodule inputs, and GitHub release workflows provide layers. Missing universal immutable action pinning, release SBOMs, signatures, provenance, and documented repository access controls leave material residual risk. |
```
JustinCappos: I assume you're missing attestations as well then, right?  These may be easier to add and serve and a more machine readable, security relevant control for provenance and signing etc.

projects/higress/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,283 @@
...
+versioned, but release artifacts do not currently have a project-generated
+SBOM, signature, or SLSA provenance. Not all workflow actions are pinned to
+immutable commits. The public repository does not prove a required reviewer
+count, signed-commit requirement, organization-wide 2FA, or branch-protection
+configuration; those controls require separate repository-settings evidence.
+
+Ordinary project-team communication uses GitHub issues, pull requests,
+discussions, localized community channels, and Discord. Inbound users use the
+same public channels. Every vulnerability report must be submitted to both
+GitHub Private Security Advisories and the Alibaba Security Response Center,
```
JustinCappos: why do you have this duality?  This is quite odd.  Does this mean that every maintainer is granted access to the Alibaba Security Response Center as well?

JustinCappos (commented):
I have a few minor comments.  Can you take a look and address?

projects/higress/security-assessment/self-assessment.md
EndlessSeeker: Thanks — updated. The Background now explains the request path and the roles of the controller and Envoy without assuming familiarity with control-plane, data-plane, or xDS terminology.

projects/higress/security-assessment/self-assessment.md
EndlessSeeker: Good catch. AI prompts are not a separate Higress component; they are request-body data when Higress is configured as an AI gateway. We updated HG-TM-07 to say that explicitly and to distinguish prompts from model responses.

projects/higress/security-assessment/self-assessment.md
EndlessSeeker: Yes. We have made that gap explicit. The assessment now states that project-wide SBOM, artifact-signature, and machine-readable provenance-attestation coverage is incomplete. This wording also avoids implying that no individual artifact has any attestation today.

projects/higress/security-assessment/self-assessment.md
EndlessSeeker: Thanks for calling this out. No, ASRC access is not granted to every maintainer. We changed the project policy in higress-group/higress#4624: GitHub Private Security Advisories are now the required, vendor-neutral, authoritative, and sufficient intake; ASRC is an optional additional channel, and ASRC access is not required for Security Response Team membership. The project-maintained assessment was updated in higress-group/community#7 and is now synchronized here.

EndlessSeeker:
Thanks @JustinCappos for the review. We have addressed all four comments in the latest revision and replied inline with the details. The corresponding project-side changes are merged in higress-group/higress#4624 and higress-group/community#7, and the assessment here is now synchronized with that source. Please let us know if any further clarification would be helpful.

One separate process note: the `label` check currently fails before reading the PR files because `actions/checkout@v5` refuses to check out fork code from a `pull_request_target` workflow. DCO passes, and this appears to require maintainer-side workflow or label handling rather than a change to this document.

JustinCappos (approved):
okay, thanks!  This looks good to me now.  

I think doing a joint assessment from where you are now would be a moderate amount of work, but would not be extremely difficult.

EndlessSeeker:
Thanks @JustinCappos! We really appreciate your review and guidance.

EndlessSeeker:
Hi @sherine-k — a quick update: Higress has now submitted its Incubation application (#2291) and all five adopter questionnaires. All review comments on this PR have been addressed, and @JustinCappos has approved the latest revision. When you have a chance, could you please take a look for the additional approval required to merge? Thanks!

projects/higress/security-assessment/self-assessment.md
```diff
@@ -0,0 +1,290 @@
...
+| Document | Location |
+| --- | --- |
+| Vulnerability reporting and response | [`SECURITY.md`](https://github.com/higress-group/higress/blob/main/SECURITY.md) |
+| Architecture | [`docs/architecture.md`](https://github.com/higress-group/higress/blob/main/docs/architecture.md) |
+| Helm defaults | [`helm/core/values.yaml`](https://github.com/higress-group/higress/blob/main/helm/core/values.yaml) |
+| OpenSSF Best Practices | [Passing](https://www.bestpractices.dev/projects/12667), verified 2026-08-04 |
+
+This is the project-maintained working copy. For formal Due Diligence, a vetted
+snapshot must be archived at the path above in `cncf/toc`. The reviewer should
+freeze evidence links to the reviewed revision when that snapshot is archived.
```
sherine-k: Nit: If I'm not mistaken, this line is from the original template, isn't relevant for higress, and can be removed

sherine-k (commented):
Thank you for this very clear self-assessment. 
On the overall, I agree with @Santoshkumarpuppala about the difficulty for the reader to clearly identify the responsibility of the operator vs that of higress for the various threats. 
This might be more explicit in the higress docs? In other words, when it's the operator's responsibility (for ex to set RBAC, limits, etc. or to validate CRs, pluginConfigs and OCI images... ): is this clearly stated ?

EndlessSeeker:
@sherine-k Thanks for pointing this out. The responsibilities were mentioned in the assessment, but they needed a clearer explanation. We have added one here and proposed a short [operator guide](https://github.com/higress-group/community/pull/11), which is awaiting merge. It explains who sets permissions and resource limits, who reviews plugin settings and images, and what Higress applies automatically. For example, accepting a configuration does not mean it is safe, and loading an image does not mean it is trusted. We also removed the working-copy note you flagged.

PR #2272, opened 2026-08-10
Add Dragonfly to I/O bottlenecks in AI Scheduling Challenges whitepaper

pmady:
Follow-up to #2164. Marlow asked me to send this straight to cncf/toc.

**Ch 4, I/O Bottlenecks:** the "Model loading for inference" list has an empty bullet right after "Cold start" looks like a placeholder that shipped blank. Filled it with the scale-out case: when a lot of replicas start at once they each pull the same weights independently and hammer the registry. The streaming tools in the cold start bullet help per-pod latency but don't do anything for that.

**Ch 5:** added Dragonfly (CNCF graduated) to Data and Storage Tools and to the I/O Bottlenecks row of the mapping table. Alphabetical in both, to match what's already there.

Happy to reword or make any changes.

PR #2281, opened 2026-08-20
Docs(scheduling-whitepaper): Add Pavan Madduri to contributors

pmady:
Follow-up to #2272 (merged by @angellk): adding myself to the Contributors list of the two chapters I contributed to, per the existing convention.

The Contributors list appears to be series-wide across all five chapters. Happy to extend this to the remaining chapter files, or drop it entirely, if the bar for the contributors list is different.


pmady:
Hi @angellk, following up on this one. It's the two-line contributors entry for the scheduling whitepaper, following on from #2272 which you merged on Aug 20. DCO is green. Is this something you can merge, or does the contributors list go through a different process?

PR #2256, opened 2026-08-01
docs(operations): adopter submission form link and TAB reference architecture substitution

angellk:
## Summary

- Links the [Adopter Submission Form](https://docs.google.com/forms/d/1n1oLC6IKj5-7S_xeEjIdEjbtS9SWniuAo7IIOyLFuK8/edit) in the triage checklist and adopter interview section of the DD guide
- Adds guidance that a published [TAB Reference Architecture](https://github.com/cncf/tab/blob/main/process/reference-architectures.md) may serve as a replacement for an individual adopter interview

## Test plan

- [x] Verify markdown renders correctly
- [x] Confirm form link resolves
- [x] Confirm TAB reference architecture link resolves

PR #2282, opened 2026-08-22
docs(operations): allow CNCF Case Study as adopter interview substitute

angellk:
## Summary

Extends the adopter interview substitution guidance added in #2256 (now merged) so that **both** a published [TAB Reference Architecture](https://github.com/cncf/tab/blob/main/process/reference-architectures.md) **and** a published [CNCF Case Study](https://github.com/cncf/foundation/blob/main/policies-guidance/case-study-guidelines.md) may serve in place of an individual adopter interview.

Adds that it is **at the TOC sponsor's discretion during the Due Diligence** to determine whether the Reference Architecture or Case Study meets the necessary standards and includes the requisite information the Due Diligence requires to determine whether the project meets the requirements of the level in process.

## Changes

- `operations/dd-toc-guide.md`: updated the substitution note in the Conducting Adopter Interviews section to cover Case Studies alongside Reference Architectures, and to state the sponsor-discretion standard.

## Test plan

- [x] Verify markdown renders correctly
- [x] Confirm the CNCF Case Study guidelines link resolves
- [x] Confirm the TAB Reference Architecture link resolves


jrsapi:
Looks good overall; I have one requested change before merge.
The existing note cautions that interviewees may be project maintainers or governance participants rather than operational users. The same issue can arise with a Reference Architecture or Case Study: it might describe the project team’s or a vendor’s work without showing an independent adopter’s operational use.
Could we add language asking the TOC member to make that same check when using either as an interview substitute?
Suggested addition: “When evaluating a Reference Architecture or Case Study, the TOC member should confirm that it demonstrates an independent organization’s operational use of the project, rather than relying solely on the project team’s or a vendor’s account.”
LGTM otherwise.

angellk:
Thanks @jrsapi - good catch. I've added your suggested language so the same independence check applies to substitutes:

> When evaluating a Reference Architecture or Case Study, the TOC member should confirm that it demonstrates an independent organization's operational use of the project, rather than relying solely on the project team's or a vendor's account.

Pushed in 527d08b. I also mirrored it to the contribute-site guide and the announcement blog (cncf/contribute-site#397) to keep the two in sync.

jrsapi (approved):
LGTM

operations/dd-toc-guide.md
```diff
@@ -282,7 +282,7 @@ It is anticipated that a minimum of three adopter interviews are required to app
...
...ct's inclusion in a published [TAB Reference Architecture](https://github.com/cncf/tab/blob/main/process/reference-architectures.md) or a published [CNCF Case Study](https://github.com/cncf/foundation/blob/main/policies-guidance/case-study-guidelines.md) may serve as a replacement for an individual adopter interview. Both represent validated, production-level adoption by the submitting organization and can provide equivalent signal to an adopter interview. When evaluating a Reference Architecture or Case Study, the TOC member should confirm that it demonstrates an independent organization's operational use of the project, rather than relying solely on the project team's or a vendor's account. It is at the TOC member's discretion who is leading the Due Diligence for Moving Levels to determine whether the Reference Architecture or Case Study meets the necessary standards and includes the requisite information the Due Diligence requires to determine whether the project meets the requirements of the level in process. Where it does, TOC members should note the Reference Architecture or Case Study in the adoption section of the Due Diligence in place of the interview summary it replaces.
```
brandtkeller: As it stands I think this is appropriate. The initial language feels more like it qualifies the content rather than being content itself. 

> This is not a change to that requirement; the TOC is clarifying the forms of validated adoption evidence that may satisfy it.

and the preceding sentence feel a bit abstract when reading the full guide in isolation. Does the reader understand why/what is being clarified - and does it need to be explicitly defined as a clarification?

angellk: Please suggest language that would help clarify!

brandtkeller: I am curious what you think about dropping those first two sentences entirely?

angellk: sgtm!

angellk: Done in 57d5d60 - dropped the first two sentences so the note opens directly with what may substitute for an interview. Thanks for the nudge, reads much cleaner.

angellk:
merging per @brandtkeller and @jrsapi lgtm

